PR #11, opened 2017-05-12
Move logging.properties from src/ to root.

wiarlawd:
All the other connectors and the library have it in the root.


PapaJoltOmega (approved):

BrettMichaelJohnson (commented):
GLTM

PR #12, opened 2017-05-12
Validate Display URLs

BrettMichaelJohnson:
This adds a simple URI validator. The validator checks that the URI
is syntactically valid, and tests the reachability of the URI's host.

### DIFF
--- a/src/com/google/enterprise/adaptor/database/ValidatedUri.java
+++ b/src/com/google/enterprise/adaptor/database/ValidatedUri.java
@@ -1,0 +1,104 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.enterprise.adaptor.database;
+
+import com.google.common.base.Strings;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Validates URIs by syntax checking and validating the host is reachable.
+ */
+public class ValidatedUri {
+  /** The logger for this class. */
+  private static final Logger log =
+      Logger.getLogger(ValidatedUri.class.getName());
+
+  /** The connect timeout, in milliseconds. */
+  private static final int TIMEOUT_MILLIS = (int) TimeUnit.SECONDS.toMillis(30);
+
+  /** The validated URI. */
+  private final URI uri;
+
+  /**
+   * Attempts to validate the given URI syntax with some more stringent checks
+   * than new URI. In this case, we're mostly trying to catch typos and obvious
+   * configuration issues.
+   *
+   * @param uriString the URI to test
+   * @throws URISyntaxException if the URL syntax is invalid
+   */
+  public ValidatedUri(String uriString) throws URISyntaxException {
+    if (Strings.isNullOrEmpty(uriString)) {
+      throw new URISyntaxException("" + uriString, "null or empty URI");
+    }
+    try {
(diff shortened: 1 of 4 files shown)


test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -263,13 +263,16 @@ public void testUrlColumnModeServesResult() throws Exception {
       String content = "from a yellow url connection comes monty python";
       writeDataToFile(testFile, content);
       ResponseGenerator resgen = ResponseGenerator.urlColumn(cfg);
-      URL testUrl = testFile.toURI().toURL();
-      ResultSet rs = makeMockUrlResultSet(testUrl);
+      // File.toURI() produces file:/path/to/temp/file, which is invalid.
```
wiarlawd, line 266: So the prod code sets uar.displayUrl to new URI(testFile.toURI().toString()) and that matches your reconstructed URI? Are you intentionally using a different construction here in the test? What about testFile.toURI().toURL().toURI()?

wiarlawd (approved):

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
wiarlawd: Never mind.

src/com/google/enterprise/adaptor/database/ValidatedUri.java
```diff
@@ -0,0 +1,104 @@
...
+   */
+  public URI getUri() {
+    return uri;
+  }
+
+  /**
+   * Checks whether the URI's host is reachable without throwing exceptions.
+   * Logs a warning if the host is not reachable.
+   */
+  public ValidatedUri logUnreachableHost() {
```
PapaJoltOmega: please delete unused method

src/com/google/enterprise/adaptor/database/ValidatedUri.java
```diff
@@ -0,0 +1,104 @@
...
+  public ValidatedUri(String uriString) throws URISyntaxException {
+    if (Strings.isNullOrEmpty(uriString)) {
+      throw new URISyntaxException("" + uriString, "null or empty URI");
+    }
+    try {
+      // Basic syntax checking, with more understandable error messages.
+      // Also ensures the URI is a URL, not a URN.
+      new URL(uriString);
+      // Advanced syntax checking, with more cryptic error messages.
+      uri = new URI(uriString);
```
PapaJoltOmega, line 48: in secmgr code we endedup accepting underscores in hostnames because some customers use them.  

i was thinking we might want to let people skip url checking via config flag.

src/com/google/enterprise/adaptor/database/ValidatedUri.java
```diff
@@ -0,0 +1,104 @@
...
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Validates URIs by syntax checking and validating the host is reachable.
+ */
+public class ValidatedUri {
```
PapaJoltOmega: can be package private

PapaJoltOmega (commented):
Thank you

src/com/google/enterprise/adaptor/database/ValidatedUri.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/database/ValidatedUri.java
BrettMichaelJohnson: No. If I have to copy this into multiple connectors, I want to maintain as few diffs as possible. Code drift got us into trouble with HtmlResponseWriter and AccumulatingDocIdPusher.

src/com/google/enterprise/adaptor/database/ValidatedUri.java
BrettMichaelJohnson: Disabling syntax checking won't work in this case.  Response.setDisplayUrl() takes a URI as a parameter. Given "http://foo_bar:8080/baz", its parsed URI is scheme: http  host: null  port: -1  path: /baz  fragment: null  query: null.  So new URI("http://foo_bar:8080/baz").toString() results in "http:///baz" - still a broken URL.

PapaJoltOmega: You're right.  the parameter to setDisplayUrl cannot have underscore in hostname.

src/com/google/enterprise/adaptor/database/ValidatedUri.java
PapaJoltOmega: But one of the benefits of not having one class in library is getting to make custom and minimal validation in each adaptor instance.

PapaJoltOmega (approved):

PR #16, opened 2017-05-16
Cleanup the logic and messages in loadResponseGenerator.

wiarlawd:
* Use Method's introspective methods rather than string manipulation
  to enumerate valid modeOfOperation values.

* Avoid checking for both bare method names in ResponseGenerator and
  fully qualified method names. They are mutually exclusive.

* Do not log messages about trying a fully qualified name before
  checking to see if it's a fully qualified name.

### DIFF
--- a/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
+++ b/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
@@ -35,6 +35,7 @@ import com.google.enterprise.adaptor.UserPrincipal;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.PreparedStatement;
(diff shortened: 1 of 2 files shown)


PapaJoltOmega (approved):
Looks great. Thank you.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -641,6 +628,23 @@ static ResponseGenerator loadResponseGeneratorInternal(Method method,
...
   }
 
+  /** Gets the valid modeOfOperation method names from ResponseGenerator. */
+  private static String getResponseGeneratorMethods() {
+    TreeSet<String> methods = new TreeSet<>();
+    for (Method method : ResponseGenerator.class.getDeclaredMethods()) {
+      int modifiers = method.getModifiers();
+      Class<?>[] parameterTypes = method.getParameterTypes();
+      if ((modifiers & Modifier.PRIVATE) != Modifier.PRIVATE
+          && (modifiers & Modifier.STATIC) == Modifier.STATIC
```
BrettMichaelJohnson, line 638: So only factory methods?

BrettMichaelJohnson (approved):
LGTM

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
wiarlawd: Yes. These are the conditions under which the existing code to find and invoke these methods succeeds.

PR #17, opened 2017-05-17
Fix "ant run" paths to logging.properties and JDBC drivers.

wiarlawd:
* Fix the reference to logging.properties, moved in commit 0c66f20.

* Add a property to build.properties to specify JDBC drivers.

Any JDBC drivers referenced by db.driverClass in
adaptor-config.properties must be on the class path at runtime,
but there's no provision for that. The commented-out lib
directory reference in adaptor.build.classpath might have been
aimed at that. To avoid requiring changes to files under version
control, switch to using a property from build.properties (as the
Documentum connector does).

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -28,15 +28,12 @@
   <property name="compile.java.bootclasspath" value="${build.bootclasspath}"/>
 
   <path id="adaptor.build.classpath">
-<!--
-    <fileset dir="${lib.dir}">
-    </fileset>
--->
     <pathelement location="${adaptor.jar}"/>
   </path>
 
   <path id="adaptor.run.classpath">
     <path refid="adaptor.build.classpath"/>
+    <pathelement path="${jdbc.classpath}"/>
   </path>
 
   <path id="cobertura.classpath">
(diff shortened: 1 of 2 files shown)


build.xml
```diff
@@ -28,15 +28,12 @@
   <property name="compile.java.bootclasspath" value="${build.bootclasspath}"/>
 
   <path id="adaptor.build.classpath">
-<!--
```
BrettMichaelJohnson, line 31: Was this expecting the developer to dump the JDBC jar file in lib? Are there other jar file in lib we might need? Is this a shortcut for lib/**/*.jar?

wiarlawd: That's my best archaeological guess. No, the only JAR file in lib is JUnit. Plus, note that this is commented out. Yes, a FileSet with no includes or excludes selects everything in the directory.

PapaJoltOmega: Putting jdbc jar in lib would make it errorprone to accidently commit that jar.  Thanks for showing this interpretation. It wasn't intended.

PapaJoltOmega (approved):

BrettMichaelJohnson (approved):
LGTM

PR #18, opened 2017-05-19
Add H2 for testing, instead of using proxies (part 1 of 3)

wiarlawd:
Part 1: Add H2 and convert 2 tests per class.

Use H2 with an in-memory database instead of using ResultSet proxies.
This lets us specify test data more naturally, e.g., you mock a BLOB
column by creating a table with a BLOB and inserting a row. It also
applies more standard JDBC constraints and behaviors to our code than
we are likely to implement in proxies (e.g., column name matching).

JdbcFixture is an unmodified copy from the Documentum connector.


BrettMichaelJohnson (approved):
LGTM

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -51,6 +56,20 @@
   @Rule
   public ExpectedException thrown = ExpectedException.none();
 
+  @Before
```
PapaJoltOmega, line 59: previous to this CL every test had its own test-localized data.

after this CL and part2 and part3 the plan is to have tables/data for the tests ready before any of them start?

not a deal breaker, just asking.

test/com/google/enterprise/adaptor/database/JdbcFixture.java
```diff
@@ -0,0 +1,66 @@
...
+package com.google.enterprise.adaptor.database;
+
+import org.h2.jdbcx.JdbcDataSource;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+/** Manages an in-memory H2 database for test purposes. */
+public class JdbcFixture {
```
PapaJoltOmega: package private?

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -247,37 +270,49 @@ public Object invoke(Object proxy, Method method, Object[] args)
...
-    String responseMsg = new String(bar.baos.toByteArray(), "US-ASCII");
-    Assert.assertEquals(content, responseMsg);
+
+    try (Statement stmt = getConnection().createStatement();
+        ResultSet rs = stmt.executeQuery("select * from data")) {
+      Assert.assertTrue("ResultSet is empty", rs.next());
+      resgen.generateResponse(rs, response);
+      String responseMsg = new String(bar.baos.toByteArray(), UTF_8);
+      Assert.assertEquals(content, responseMsg);
+    }
```
PapaJoltOmega, line 292: should tests drop what they create?  curious.

PapaJoltOmega (commented):
Thank you

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
wiarlawd: That isn't entirely true, I'm actually moving some test data out of the proxies and into the tests. The plan is to have the test data local to each test. I'm experimenting a little with the tables. If they are globally obvious (like ACLs) I put them in `@Before`. If they are intrinsically tied to the test data, as with most of the tests, I'm keeping the tables local as well.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
wiarlawd: I can't tell if you're suggesting that they should or shouldn't. We've always deleted the data in an `@After` method like I'm doing here, to reduce the complexity in the tests and make it impossible to forget.

test/com/google/enterprise/adaptor/database/JdbcFixture.java
wiarlawd: Done

PapaJoltOmega (approved):

PR #21, opened 2017-05-20
Clean up Assert, charsets, array decls, and test failures.

wiarlawd:
* Statically import Assert methods.
* Use UTF_8. For CLOBs, the connector writes them as UTF-8, so we
  should read them as such. For BLOBs, it doesn't matter much, because
  we're starting with strings. US-ASCII and UTF-8 work the same on
  7-bit bytes, but US-ASCII converts high-bit bytes to ?, so UTF-8 is
  a little safer.
* Use ByteArrayOutputStream.toString to avoid an extra array copy.
* Fix array declarations per the style guide ([] follows the type).
* Fix a test failure from an unlucky merge in commit 84540f9.


BrettMichaelJohnson (approved):
Any idea how the display url merge failed?

wiarlawd:
You edited all the existing URIs, but commit d4dc983 added a new one and was committed before yours.

PR #22, opened 2017-05-21
Convert remaining tests to H2 (part 3 of 3).

wiarlawd:


test/com/google/enterprise/adaptor/database/JdbcFixture.java
```diff
@@ -48,9 +54,34 @@ public static Connection getConnection() {
   }
 
   public static void dropAllObjects() throws SQLException {
+    try {
+      for (AutoCloseable object : openObjects) {
+        object.close();
+      }
+    } catch (Exception e) {
```
PapaJoltOmega, line 61: so first error means we don't try to close later objects?

PapaJoltOmega (approved):
Is H2 in these tests reliable?  That is, what is the flakiness rate?

It's not clear to me what payoff we get for all these test changes.

Thank you

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -382,14 +240,20 @@ public void testFilepathColumnModeServesResult() throws Exception {
         Response.class.getClassLoader(),
         new Class[] { Response.class }, far);
     Map<String, String> cfg = new TreeMap<String, String>();
-    cfg.put("columnName", "my-filepath-col");
+    cfg.put("columnName", "filepath");
     File testFile = null;
     try {
```
BrettMichaelJohnson, line 245: Can we convert these to use JUnit temp files so we can dispense with all the try/finally blocks?

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -460,15 +334,18 @@ public void testUrlColumnModeIncorrectColumnName() throws Exception {
...
+    cfg.put("columnName", "wrongcolumn");
     File testFile = null;
     try {
       testFile = File.createTempFile("db.rg.test", ".txt");
       String content = "from a yellow url connection comes monty python";
       writeDataToFile(testFile, content);
+
+      executeUpdate("create table data(url varchar)");
+      executeUpdate("insert into data(url) values ('some URL')");
+
```
BrettMichaelJohnson, line 346: "some URL"?  You really didn't need to write out a temp file.

test/com/google/enterprise/adaptor/database/JdbcFixture.java
wiarlawd: Yes. I thought about it, but the chances of a problem are rare, and likely point out flaws in the test.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
wiarlawd: Ah, good catch. Yes, I started inserting the real URI into the database, but realized that the test shouldn't try to read the real URL column, so making it an invalid URI seemed good. The test is good, the file is just redundant, so let me take a rain check and include this as part of the followup cleanup of the temp files.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
wiarlawd: Yes, I planned to do that in a followup CL. I wanted to keep this focused on the H2 change.

wiarlawd (commented):
We've been using H2 in the tests for about 8 years and have never seen a flake.

As I mentioned in the part 1 CL, the two main benefits are the naturalness of adding tests, especially for specific data types, and secondly the fact that we're actually using a JDBC driver, which has more complicated behaviors that the connector has to deal with, but which are rarely implemented in the proxies. For example, we're missing a lot of data type tests, and there are a handful of bugs in the data type handling, and we're not mocking the JDBC semantics of column lookup, and there are bugs for the column lookup. We're also not calling getDocIds, getModifiedDocIds, or getDocContent at all in the tests, and part of the reason is that they create a JDBC connection based on the config, and that can't be mocked without a lot more effort. But with H2 it's trivial to configure the adaptor to run against H2, which lets us write more tests.

BrettMichaelJohnson (approved):
LGTM

PapaJoltOmega (approved):

PR #23, opened 2017-05-21
Use a more realistic query for UniqueKey round trip test.

wiarlawd:
The query in the test should mimic db.singleDocContentSql and
successfully fetch the row the unique key was generated from.


BrettMichaelJohnson (approved):
LGTM.  I didn't understand the purpose of deleting, then re-inserting the data in the original test.

PR #28, opened 2017-05-24
Include the resources in the class path for "ant run".

wiarlawd:
With this, modeOfOperation=rowToHtml fails with
"Default stylesheet not found in resources".


BrettMichaelJohnson (approved):
s/With this/Without this/

PR #34, opened 2017-05-26
Make MetadataColumns a Map

BrettMichaelJohnson:


src/com/google/enterprise/adaptor/database/MetadataColumns.java
```diff
@@ -28,7 +29,7 @@
  * Stores mapping of db columns to metadata keys to send.
  * This class is thread-safe.
  */
-class MetadataColumns {
+class MetadataColumns extends AbstractMap {
```
wiarlawd: Does this really compile without warnings? I think it should be AbstractMap<String, String>.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -476,11 +476,11 @@ public void testIncludeAllColumnsAsMetadata_mcSet() throws Exception {
     final Config config = createStandardConfig(moreEntries);
     DatabaseAdaptor adaptor = new DatabaseAdaptor();
     adaptor.init(TestHelper.createConfigAdaptorContext(config));
-    assertEquals("MetadataColumns({col2=gsa2, db_col1=gsa1})",
+    assertEquals("{col2=gsa2, db_col1=gsa1}",
```
wiarlawd: Not nearly as much of this as a I thought. Since we're not using Guava, what do you think about using something like this:

  @SuppressWarnings("unchecked")
  private <T> Map<T, T> asMap(T... entries) {
    if (entries.length % 2 != 0) {
      throw new IllegalArgumentException(
          "A map requires an even number of entries");
    }
    Map<T, T> map = new HashMap<T, T>();
    for (int i = 0; i < entries.length - 1; i += 2) {
      map.put(entries[i], entries[i + 1]);
    }
    return map;
  }

wiarlawd (approved):
LGTM, with a nit and suggestion.

src/com/google/enterprise/adaptor/database/MetadataColumns.java
```diff
@@ -19,6 +19,7 @@
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.util.Collections;
+import java.util.AbstractMap;
```
wiarlawd: Wrong order

wiarlawd (changes requested):
Lint error in your previous submit (assertFalse is unused in DBAdatptorTest), and new one here.

src/com/google/enterprise/adaptor/database/MetadataColumns.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/database/MetadataColumns.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: Done.

BrettMichaelJohnson (commented):
Also fixed unused imports

wiarlawd (approved):

PR #35, opened 2017-05-27
Fix b/18112595 Support Add/Delete Action Column

BrettMichaelJohnson:


test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -675,6 +677,184 @@ public void testGetDocIds() throws Exception {
   }
 
   @Test
+  public void testGetDocIdsActionColumnDelete() throws Exception {
```
wiarlawd: It seems reasonable to me to have four rows in the table instead of four tests, with the benefit of testing to make sure the actions are truly row-based.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -675,6 +677,184 @@ public void testGetDocIds() throws Exception {
...
+    configEntries.put("db.password", "");
+    configEntries.put("db.url", JdbcFixture.URL);
+    configEntries.put("db.uniqueKey", "id:int");
+    configEntries.put("db.everyDocIdSql", "select * from data");
+    configEntries.put("db.singleDocContentSql", "");
+    configEntries.put("db.singleDocContentSqlParameters", "");
+    configEntries.put("db.aclSqlParameters", "id");
+    configEntries.put("adaptor.namespace", "Default");
+    configEntries.put("db.modeOfOperation", "urlAndMetadataLister");
+    configEntries.put("db.modeOfOperation.urlAndMetadataLister.columnName",
```
wiarlawd: Even in our twisted config world, some of these are weird. This one, for example, doesn't exist and logs a warning if you use it.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -675,6 +677,184 @@ public void testGetDocIds() throws Exception {
   }
 
   @Test
+  public void testGetDocIdsActionColumnDelete() throws Exception {
+    executeUpdate("create table data(id integer, action varchar)");
+    executeUpdate("insert into data(id, action) values('1001', 'delete')");
+
+    Map<String, String> configEntries = new HashMap<String, String>();
+    configEntries.put("db.actionColumn", "action");
```
wiarlawd: We don't put something first just because it's under test elsewhere. This should be cleaned up, but either after uniqueKey or before metadataColumns makes more sense to me, especially the latter.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -260,6 +270,14 @@ public void getDocIds(DocIdPusher pusher) throws IOException,
     outstream.forcePush();
   }
 
+  private boolean isDeleteAction(ResultSet rs) throws SQLException {
+    if (!actionColumn.equals("")) {
+      String action = rs.getString(actionColumn);
+      return (action != null && "delete".equals(action.toLowerCase(US)));
```
wiarlawd, line 276: I'm not actually sure whether equalsIgnoreCase is better or worse than this.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: In another CL, I address the idiotic default values of createStandardConfig. Consequently, all the remaining config entries here go away, with the exception of this one. Oops. Just noticed I had forgotten to make a PullRequest from the branch.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: Just a copy/paste from Srinivas' test.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: done.

wiarlawd (approved):

PR #39, opened 2017-05-27
Verify all column names in init, if possible (b/29425538).

wiarlawd:
This is a best effort, and SQLExceptions are not fatal. Even verified
column names might fail at runtime in certain circumstances.

Related changes:
* Suppress verification in many of the tests. We could revisit this
  after fixing the test configs. but the required config may obscure
  the actual point of these tests.

Other changes:
* Remove db.modeOfOperation.urlAndMetadataLister.columnName config.

### DIFF
--- a/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
+++ b/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
@@ -542,15 +542,15 @@ public class DatabaseAdaptorTest {
 
   @Test
   public void testInitVerifyColumnNames_urlAndMetadata() throws Exception {
-    executeUpdate("create table data(id int, other varchar)");
+    executeUpdate("create table data(url varchar, other varchar)");
 
     Map<String, String> moreEntries = new HashMap<String, String>();
     moreEntries.put("db.modeOfOperation", "urlAndMetadataLister");
     moreEntries.put("docId.isUrl", "true");
-    moreEntries.put("db.uniqueKey", "id:int");
+    moreEntries.put("db.uniqueKey", "url:string");
     moreEntries.put("db.singleDocContentSqlParameters", "");
     moreEntries.put("db.aclSqlParameters", "");
-    moreEntries.put("db.everyDocIdSql", "select id from data");
+    moreEntries.put("db.everyDocIdSql", "select url from data");
     moreEntries.put("db.singleDocContentSql", "");
     moreEntries.put("db.metadataColumns", "other:other");
     moreEntries.put("adaptor.namespace", "Default");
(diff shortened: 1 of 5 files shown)


test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -177,17 +222,14 @@ public void testInitOfUrlMetadataListerDocIdIsUrlTrue() throws Exception {
...
-    configEntries.put("db.url", "must be set");
-    configEntries.put("db.user", "must be set");
-    configEntries.put("db.password", "must be set");
+    configEntries.put("db.driverClass", JdbcFixture.DRIVER_CLASS);
+    configEntries.put("db.url", JdbcFixture.URL);
+    configEntries.put("db.user", JdbcFixture.USER);
+    configEntries.put("db.password", JdbcFixture.PASSWORD);
     configEntries.put("db.uniqueKey", "must be set:string");
-    configEntries.put("db.everyDocIdSql", "must be set");
+    configEntries.put("db.everyDocIdSql", "select 42 as \"must be set\"");
```
BrettMichaelJohnson, line 234: I suggest we compromise on our two approaches. Set all these (must be set)  to empty string if that will cause them to pass your validation.  Any test will then need to set uniqueId, queries, etc as needed to function.  That also avoids setting them to empty string in all the tests where you added it.

wiarlawd: If we exclude the connection properties, there are only 4 required properties (uniqueKey, everyDocIdSql, singleDocContentSql, and modeOfOperation). I don't think it's unreasonable for each test to set those. I am interested in getting all of the tests to the point where we don't have to specify the default values, but I don't know if I want to go past that. I do, as I mention in the commit message, want to revisit the suppressions once we get there. I started fixing them so the tests pass with them, but it was a complicated can of worms, and I wanted to rethink it once each test was only specifying the necessary properties.

BrettMichaelJohnson (approved):

wiarlawd:
PJ, I just wanted you to nod at the idea, though you could look at the code as desired.

Basically, any SQLException will skip the related verification. We do a case-insensitive comparison to match the column names, so "id", "Id", and "ID" all match. If the database is case-sensitive, that may be not true when the queries are really run. However, we specifically avoid detecting a possible case-sensitivity and verifying that. If a result set has "id" and "ID" columns, we will allow a reference to "Id" to proceed, and perhaps succeed and perhaps fail, depending on the JDBC driver and DBMS.

PapaJoltOmega (approved):
Thank you. LGTM

wiarlawd:
Brett, PTAL. This was a non-trivial merge to fix tests broken by earlier commits.

BrettMichaelJohnson (approved):

PR #40, opened 2017-05-27
UniqueKey can be case insensitive looking up column names and types.

BrettMichaelJohnson:
This change allows the contentSqlColumns and aclSqlColumns column names
to match any of the ukDecls case insensitively, IFF all the ukDecls can
be uniquely lowercased.

### DIFF
--- a/src/com/google/enterprise/adaptor/database/UniqueKey.java
+++ b/src/com/google/enterprise/adaptor/database/UniqueKey.java
@@ -14,6 +14,8 @@
 
 package com.google.enterprise.adaptor.database;
 
+import static java.util.Locale.US;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.enterprise.adaptor.InvalidConfigurationException;
 
@@ -49,6 +51,8 @@ class UniqueKey {
   private final List<String> contentSqlCols;  // columns for content query
   private final List<String> aclSqlCols;  // columns for acl query
 
+  private boolean caseSensitiveKeys; // if types map has case sensitive keys
+
   UniqueKey(String ukDecls, String contentSqlColumns, String aclSqlColumns) {
     if (null == ukDecls) {
       throw new NullPointerException();
@@ -81,24 +85,24 @@ class UniqueKey {
       def[1] = def[1].trim();
       String name = def[0];
       ColumnType type;
-      if ("int".equals(def[1].toLowerCase())) {
+      if ("int".equals(def[1].toLowerCase(US))) {
         type = ColumnType.INT;
-      } else if ("string".equals(def[1].toLowerCase())) {
+      } else if ("string".equals(def[1].toLowerCase(US))) {
         type = ColumnType.STRING;
-      } else if ("timestamp".equals(def[1].toLowerCase())) {
+      } else if ("timestamp".equals(def[1].toLowerCase(US))) {
         type = ColumnType.TIMESTAMP;
-      } else if ("date".equals(def[1].toLowerCase())) {
+      } else if ("date".equals(def[1].toLowerCase(US))) {
         type = ColumnType.DATE;
-      } else if ("time".equals(def[1].toLowerCase())) {
+      } else if ("time".equals(def[1].toLowerCase(US))) {
         type = ColumnType.TIME;
-      } else if ("long".equals(def[1].toLowerCase())) {
+      } else if ("long".equals(def[1].toLowerCase(US))) {
         type = ColumnType.LONG;
       } else {
         String errmsg = "Invalid UniqueKey type '" + def[1] + "'. Valid"
             + " types are: int, string, timestamp, date, time, and long.";
         throw new InvalidConfigurationException(errmsg);
       }
-      if (tmpTypes.containsKey(name)) {
+      if (tmpNames.contains(name)) {
         String errmsg = "Invalid db.uniqueKey configuration: key name '"
             + name + "' was repeated.";
         throw new InvalidConfigurationException(errmsg);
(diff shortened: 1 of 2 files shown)


test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -104,6 +104,32 @@ public void testNameRepeatsNotAllowed() {
...
+    assertEquals("num", uk.nameForTest(0));
+    assertEquals("Num", uk.nameForTest(1));
+    assertEquals(UniqueKey.ColumnType.INT, uk.typeForTest(0));
+    assertEquals(UniqueKey.ColumnType.STRING, uk.typeForTest(1));
+  }
+
+  @Test
+  public void testCaseInsensitiveNameNotAllowedIfCaseSensitiveKeys() {
+    thrown.expect(InvalidConfigurationException.class);
+    UniqueKey uk = new UniqueKey("id:int,Id:string", "ID", "ID");
```
wiarlawd: I mentioned this week that I had explicitly rejected this logic. If the database has case-sensitive column names, I don't think we can predict whether a lookup with yet another case will succeed or not.

I haven't thought this completely through, but I think if you make the tmpTypes keys case-insensitive, with the String.CASE_INSENSITIVE_ORDER, you might not have to do anything else.

wiarlawd (changes requested):

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
wiarlawd: A little more complicated than that, since you have to maintain the original map for the types.

wiarlawd:
I don't know if you're starting completely over here, too, but please include b/38283781 in the title.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -107,28 +111,38 @@
       tmpTypes.put(name, type);
     }
     names = Collections.unmodifiableList(tmpNames);
-    types = Collections.unmodifiableMap(tmpTypes);
+    Map<String, ColumnType> caseInsensitiveTypes = new TreeMap<>();
```
wiarlawd: What about using TreeMap<>(Strings.CASE_INSENSITIVE_ORDER) here, as verifyColumnNames does, then just caseInsensitiveTypes.putAll(tmpTypes), and if the sizes differ, types = tmpTypes, otherwise caseInsensitiveTypes. Then we can drop the caseSensitive field, and put splitIntoNameList back to static (as preferred when called from a constructor).

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
wiarlawd: As discussed, the problem arises with mixed case that we don't know how to lookup the type of "ID". We're not requiring any particular behavior of the database, we're just saying that if the database is case-sensitive and your unique key exposes that via duplicates, that your parameters config much be case-sensitive as well.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -104,6 +104,32 @@ public void testNameRepeatsNotAllowed() {
   }
 
   @Test
+  public void testCaseSensitiveNameRepeatsAllowed() {
+    UniqueKey uk = new UniqueKey("num:int,Num:string");
+    assertEquals(2, uk.numElementsForTest());
+    assertEquals("num", uk.nameForTest(0));
+    assertEquals("Num", uk.nameForTest(1));
+    assertEquals(UniqueKey.ColumnType.INT, uk.typeForTest(0));
```
wiarlawd: As a separate CL, what would you think about using H2 with prepared statements to test things like this through the public interface, rather than having all these test-specific methods in UniqueKey?

wiarlawd (commented):
And now I'll review the new commit.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -82,27 +84,35 @@
...
       }
-      if (tmpTypes.containsKey(name)) {
+      if (tmpNames.contains(name)) {
         String errmsg = "Invalid db.uniqueKey configuration: key name '"
             + name + "' was repeated.";
         throw new InvalidConfigurationException(errmsg);
+      } else if (tmpTypes.containsKey(name) && tmpTypes.get(name) != type) {
+        // The ukDecls contain two keys that differ only in case but are of
+        // different types. Force a case-sensitive type lookup.
+        Map<String, ColumnType> caseSensitiveTypes = new TreeMap<>();
```
wiarlawd: OK, we're inside a for loop here. This is hurting my head a little. So you're going to swap a partial tmpTypes into new case-sensitive mapping, and then continue building the types. Maybe a comment to that effect: "Replace the partial map with a case-sensitive copy."

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -82,27 +84,35 @@
       def[1] = def[1].trim();
       String name = def[0];
       ColumnType type;
-      if ("int".equals(def[1].toLowerCase())) {
+      if ("int".equals(def[1].toLowerCase(US))) {
```
wiarlawd: Getting uglier. Maybe switch (def[1].toLowerCase(US)) { case "int": ... }. Or perhaps better, ColumnType.valueOf(def[1].toUpperCase()), with a less helpful error message, or catch the IllegalArgumentException to produce the better one we have now.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -316,7 +326,13 @@ String nameForTest(int i) {
   /** Type of particular column in primary key. */
   @VisibleForTesting
   ColumnType typeForTest(int i) {
-    return types.get(names.get(i));
+    return getType(names.get(i));
+  }
+
+  /** Type of a named column in primary key. */
+  @VisibleForTesting
+  ColumnType getType(String name) {
```
wiarlawd: Why not a typeForTest(String) overload?

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -119,23 +129,23 @@
...
       aclSqlCols = names;
     } else {
-      aclSqlCols = splitIntoNameList(aclSqlColumns, tmpTypes.keySet());
+      aclSqlCols = splitIntoNameList(aclSqlColumns);
     }
   }
 
-  private static List<String> splitIntoNameList(String cols,
-      Set<String> validNames) {
+  private List<String> splitIntoNameList(String cols) {
```
wiarlawd: Now I would certainly undo the s/static// and s/types.get/getType/ changes below.

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -82,27 +85,26 @@
...
       }
-      if (tmpTypes.containsKey(name)) {
+      if (tmpNames.contains(name)) {
         String errmsg = "Invalid db.uniqueKey configuration: key name '"
             + name + "' was repeated.";
         throw new InvalidConfigurationException(errmsg);
+      } else if (tmpTypes.containsKey(name) && tmpTypes.get(name) != type) {
+        // The ukDecls contain two keys that differ only in case but are of
+        // different types. Force a case-sensitive type lookup by replacing
+        // the case-insensitive type map with a case-sensitive one.
```
wiarlawd: The "partial" and "copy" are important to me. I would change this line to: "the partial type map with a case-sensitive copy."

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -82,27 +85,26 @@
...
+        Map<String, ColumnType> caseSensitiveTypes = new TreeMap<>();
```
wiarlawd: Please add a test that requires this copy-from-scratch, e.g., "id:int,ID:int,Id:string", or even just change testCaseSensitiveNameRepeatsAllowed to "num:int,nuM:int,Num:string"

src/com/google/enterprise/adaptor/database/UniqueKey.java
BrettMichaelJohnson: done.

src/com/google/enterprise/adaptor/database/UniqueKey.java

wiarlawd (approved):

PR #42, opened 2017-05-31
Fix b/8433358. Support timestamp field.

srinicodebytes:


src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -378,14 +378,23 @@ private void addMetadata(MetadataHandler meta, ResultSet rs)
             }
           }
           break;
+        case Types.DATE:
+          Date dt = rs.getDate(index);
+          value = dt.toString();
```
wiarlawd: Handle null, here and below, and add tests for the same (for each data type, since you insist on forgetting it so often).

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -378,14 +378,23 @@ private void addMetadata(MetadataHandler meta, ResultSet rs)
             }
           }
           break;
+        case Types.DATE:
+          Date dt = rs.getDate(index);
```
wiarlawd: s/Date/java.sql.Date/

This is currently a java.util.Date. No real badness, just a loss of clarity and temptation to misuse. Note that you can't import this, because we're importing java.util.Date.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -926,4 +926,95 @@ public void testClobColumnWithNullMetadata() throws Exception {
     expected.add("col2", "null");
     assertEquals(expected, response.getMetadata());
   }
+
+  @Test
+  public void testColumnTypeDate() throws Exception {
+    executeUpdate("create table data(id  integer, col  date)");
```
wiarlawd: Same double spaces.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -926,4 +926,95 @@ public void testClobColumnWithNullMetadata() throws Exception {
...
+    executeUpdate("insert into data(id, col) values('1001', {d '2004-10-06'})");
```
wiarlawd: 80 cols

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -926,4 +926,95 @@ public void testClobColumnWithNullMetadata() throws Exception {
     expected.add("col2", "null");
     assertEquals(expected, response.getMetadata());
   }
+
+  @Test
+  public void testColumnTypeDate() throws Exception {
```
wiarlawd: These names don't match the earlier CLOB test names. Let's have a better scheme, like testMetadataColumns_date and testMetadataColumns_dateNull. Rename the CLOB tests as well.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
wiarlawd: Don't quote integers in SQL. It's not actually valid SQL, so it's not portable. We don't care much about portability, but it's just a bad habit.

wiarlawd (changes requested):
This shouldn't be Oracle-specific. Please try SQL Server, and remove "Oracle" from the title when this is committed.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -926,4 +926,185 @@ public void testClobColumnWithNullMetadata() throws Exception {
     expected.add("col2", "null");
     assertEquals(expected, response.getMetadata());
   }
+
+  @Test
+  public void testMetadataColumns_date() throws Exception {
+    executeUpdate("create table data(id  integer, col date)");
```
wiarlawd: Missed comment: delete extra space, here and below (5 times in all).

wiarlawd (changes requested):
Go ahead and merge in master (there are no conflicts, so you'll need to do this from git). You will get test failures, and you can also delete the JDBC connection properties from your tests.

The test failures are mostly that you need to set db.uniqueKey, and set empty     db.aclSqlParameters and db.singleDocContentSqlParameters. I also missed this earlier, but please use moreEntries in the new tests instead of configEntries. You copied configEntries from the wrong place, I think. All the actual tests were using moreEntries. It might all change soon, but let's use the most consistent thing for now. Don't bother changing the CLOB tests.

wiarlawd (approved):
LGTM

PR #43, opened 2017-05-31
Support simple identity column mappings in MetadataColumns config.

BrettMichaelJohnson:


PapaJoltOmega (approved):

src/com/google/enterprise/adaptor/database/MetadataColumns.java
```diff
@@ -62,13 +60,14 @@
       log.fine("element: `" + e + "'");
       e = e.trim(); 
       String def[] = e.split(":", 2);
-      if (2 != def.length) {
-        String errmsg = "expected two parts separated by colon: `" + e + "'";
-        throw new InvalidConfigurationException(errmsg);
+      if (def.length == 1) {
+        String columnName = def[0].trim();
+        if (!columnName.isEmpty()) {
```
wiarlawd: I don't like quietly dropping invalid entries on the floor. You're also not checking for empty strings in the else, so the behavior is inconsistent.

If we let empty names through here, as we were doing before, verifyColumnNames will now catch empty column names, but nothing will catch empty metadata names, it looks like those will flow throw to the GSA (possibly a bug in the library's Metadata class).

src/com/google/enterprise/adaptor/database/MetadataColumns.java
```diff
@@ -62,13 +60,14 @@
       log.fine("element: `" + e + "'");
       e = e.trim(); 
       String def[] = e.split(":", 2);
-      if (2 != def.length) {
-        String errmsg = "expected two parts separated by colon: `" + e + "'";
-        throw new InvalidConfigurationException(errmsg);
+      if (def.length == 1) {
+        String columnName = def[0].trim();
```
wiarlawd: We know there's at least one (possibly empty) entry in def, so you could lift these lines out of this if statement.

wiarlawd (commented):
You should file an FR for this, for QA and doc tracking.

src/com/google/enterprise/adaptor/database/MetadataColumns.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/database/MetadataColumns.java
BrettMichaelJohnson: As discussed, if def.length is 1 and columnName is empty, we got an empty entry like "foo,,bar".  The splitter on ',' already drops empty entries from the end.  This drops empty entries in the middle.

As far as invalid def.length == 2 entries, the columnName and/or the metadataKey would be empty.  I will treat an empty metadataKey as an identity mapping ("foo:" is the same as "foo"). An empty columnName here will get caught by verifyColumns in init().

wiarlawd (approved):

PR #53, opened 2017-06-02
Use try-with-resources and canonical getters in TupleReader

srinicodebytes:
- Check for null
- Use try-with-resources
- Use right JDBC getXXX methods

### DIFF
--- a/src/com/google/enterprise/adaptor/database/TupleReader.java
+++ b/src/com/google/enterprise/adaptor/database/TupleReader.java
@@ -26,6 +26,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.io.Writer;
+import java.sql.Blob;
+import java.sql.Clob;
+import java.sql.NClob;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
@@ -36,6 +39,7 @@ import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import javax.xml.bind.DatatypeConverter;
@@ -138,7 +142,6 @@ class TupleReader extends XMLFilterImpl implements XMLReader {
     ResultSetMetaData rsmd = resultSet.getMetaData();
     int colCount = rsmd.getColumnCount();
 
-    Writer outWriter = new BufferedWriter(new ContentHandlerWriter(handler));
     for (int col = 1; col <= colCount; col++) {
       atts.clear();
       String columnName = rsmd.getColumnLabel(col);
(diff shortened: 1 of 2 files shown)


src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -205,45 +289,53 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
-          reader.close();
+          if (string != null) {
+            try (Writer writer =
+                new BufferedWriter(new ContentHandlerWriter(handler))) {
+              handler.startElement("", columnName, columnName, atts);
+              writer.write(string);
+            }
+          }
           break;
         case Types.NCHAR:
```
wiarlawd: Maybe we should be calling getNString on these first two, but we're using getString elsewhere, and I think we should do the same here (especially since we changed CHAR/VARCHAR to avoid using a stream).

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -205,45 +289,53 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
+              writer.write(String.valueOf(value));
+            }
+          }
           break;
         case Types.OTHER:
         default:
           string = resultSet.getString(col);
           if (null == string) {
             string = "" + resultSet.getObject(col);
+          } else {
```
wiarlawd: What? I think calling getObject is silly here, but you're changing the behavior of null here, too, but with strange code. Let's figure out the null handling, and then fix this.

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -151,49 +154,130 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
       if (resultSet.getObject(col) == null) {
         atts.addAttribute("", "ISNULL", "ISNULL", "NMTOKEN", "true");
         handler.startElement("", columnName, columnName, atts);
-        // TODO: Re-examine this need for flush.
-        outWriter.flush();
```
wiarlawd: Why did you remove the calls to flush?

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -151,49 +154,130 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
         handler.endElement("", columnName, columnName);
         continue;
       }
       switch (sqlType) {
         case Types.DATE:
           Date dateValue = resultSet.getDate(col);
-          handler.startElement("", columnName, columnName, atts);
-          outWriter.write(DATEFMT.format(dateValue));
+          if (dateValue != null) {
+            try (Writer writer =
```
wiarlawd: s/writer/outWriter/ (not enough reason to change even more lines beyond whitespace, which can be ignored by git blame).

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -151,49 +154,130 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
-        outWriter.flush();
         handler.endElement("", columnName, columnName);
         continue;
       }
       switch (sqlType) {
         case Types.DATE:
           Date dateValue = resultSet.getDate(col);
-          handler.startElement("", columnName, columnName, atts);
-          outWriter.write(DATEFMT.format(dateValue));
+          if (dateValue != null) {
```
wiarlawd, line 158: So you're changing the null handling here. Are there any tests of it? What does the change look like in the output? Are you delaying the construction of the writer just because the null handling is more localized?

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -205,45 +289,53 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
-          Reader nReader = resultSet.getNCharacterStream(col);
-          handler.startElement("", columnName, columnName, atts);
-          copyValidXMLCharacters(nReader, outWriter);
-          nReader.close();
+          try (Reader reader = resultSet.getNCharacterStream(col);
+              Writer writer =
+                  new BufferedWriter(new ContentHandlerWriter(handler))) {
+            if (reader != null) {
+              handler.startElement("", columnName, columnName, atts);
+              copyValidXMLCharacters(reader, writer);
```
wiarlawd: Another weirdness here, and maybe this is related to Lavanya's fix, is that we were using this copyValidXMLCharacters for streams, but we're not using it for strings. I think we should split out the *CHAR types, and use getString, but wrap them in a StringReader and call copyValidXMLCharacters. (We don't need it for the integer types, which can't have invalid XML characters to begin with.)

wiarlawd (changes requested):
Mostly concerned about nulls, and the corresponding lack of test changes, indicating we probably need more tests.

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -148,52 +152,127 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
+            if (lob != null) {
+              // so encode using Base64
+              atts.addAttribute("", "encoding", "encoding", "NMTOKEN",
+                  "base64binary");
+              handler.startElement("", columnName, columnName, atts);
+              encode(lob, outWriter);
+            }
+          }
+          break;
+        case Types.LONGVARCHAR:
```
wiarlawd: Put this back down where it was, after VARCHAR.

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -202,48 +281,68 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
+              handler.startElement("", columnName, columnName, atts);
+              outWriter.write(string);
+            }
+          }
           break;
+        case Types.CHAR:
+        case Types.VARCHAR:
         case Types.NCHAR:
-        case Types.NCLOB:
+          String str = resultSet.getString(col);
```
wiarlawd: s/String str/string/, and the cascading diffs below.

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -202,48 +281,68 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
+            }
+          }
+          break;
         case Types.NVARCHAR:
         case Types.LONGNVARCHAR:
-          Reader nReader = resultSet.getNCharacterStream(col);
-          handler.startElement("", columnName, columnName, atts);
-          copyValidXMLCharacters(nReader, outWriter);
-          nReader.close();
+          try (Reader reader = resultSet.getNCharacterStream(col);
```
wiarlawd: s/reader/nReader/ here, but not in NCLOB.

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -202,48 +281,68 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
+          String str = resultSet.getString(col);
+          if (str != null) {
+            try (Writer outWriter =
+                new BufferedWriter(new ContentHandlerWriter(handler))) {
+              handler.startElement("", columnName, columnName, atts);
+              copyValidXMLCharacters(new StringReader(str), outWriter);
+            }
+          }
+          break;
         case Types.NVARCHAR:
```
wiarlawd: Move this to the previous case.

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -202,48 +281,68 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
+              outWriter.write(String.valueOf(value));
+            }
+          }
           break;
         case Types.OTHER:
         default:
-          string = resultSet.getString(col);
-          if (null == string) {
-            string = "" + resultSet.getObject(col);
+          str = resultSet.getString(col);
```
wiarlawd: s/str/string/ throughout this case.

src/com/google/enterprise/adaptor/database/TupleReader.java
BrettMichaelJohnson: Also related to null checking, the check sqlTypeName != null 3 lines up is redundant.  It cannot be null here. (FindBugs)

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -118,6 +118,24 @@ public void testVARCHAR() throws Exception {
     assertEquals(golden, result);
   }
 
+  @Test
+  public void testNullVARCHAR() throws Exception {
```
wiarlawd: testVarchar_null, here and throughout, but don't change the existing test names.

wiarlawd: This test case already existed (as testNULL), so you might as well delete testNULL.

BrettMichaelJohnson (commented):
Other issues flagged by FindBugs:

TupleReader redundantly implements XMLReader (already implemented by XMLFIlterImpl).

The SimpleDateFormats (DATEFORMAT, TIMEFORMAT, TIMEZONEFORMAT) are not thread-safe.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -118,6 +118,24 @@ public void testVARCHAR() throws Exception {
...
+    executeUpdate("create table data(id integer, colname varchar)");
```
wiarlawd: What's your reasoning for wanting the ID columns? Having a test with multiple columns seems good, as does a test with mixed null/non-null values, but it doesn't need to be replicated for each type. I would drop the ID columns in these null tests and add a single test with a few columns of duplicate and different types, with some nulls mixed in.

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -148,52 +152,127 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
+            } finally {
+              try {
+                nclob.free();
+              } catch (Exception e) {
+                log.log(Level.FINEST, "Error closing NCLOB", e);
+              }
+            }
+          }
+          break;
         case Types.VARBINARY:
```
wiarlawd: Add BINARY.

wiarlawd (changes requested):
Also, add test pairs (non-null/null) for CLOB, BINARY, and BOOLEAN.

wiarlawd:
FYI, I do not want to fix the FindBugs errors in this CL. Two of them are not very interesting, and the thread-safety deserves its own CL.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -213,6 +190,43 @@ public void testNullINTEGER() throws Exception {
   }
 
   /**
+   * Test boolean column.
+   */
+  @Test
+  public void testBOOLEAN() throws Exception {
```
wiarlawd: testBoolean

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -213,6 +190,43 @@ public void testNullINTEGER() throws Exception {
...
+        + "</table_rec>"
+        + "</table>"
+        + "</database>";
+    ResultSet rs = executeQueryAndNext("select * from data");
+    String result = generateXml(rs);
+    assertEquals(golden, result);
+  }
+
+  @Test
+  public void testNullBOOLEAN() throws Exception {
```
wiarlawd: testBoolean_null

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -158,13 +137,12 @@ public void testChar() throws Exception {
 
   @Test
   public void testNullChar() throws Exception {
```
wiarlawd: testChar_null, that's what I meant by "throughout"

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -428,4 +438,86 @@ public void testNullBLOB() throws Exception {
     String result = generateXml(rs);
     assertEquals(golden, result);
   }
+
+  /**
+   * Test CLOB column.
+   */
+  @Test
+  public void testCLOB() throws Exception {
```
wiarlawd: testClob

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -428,4 +438,86 @@ public void testNullBLOB() throws Exception {
...
+            + " would consider on a case-by-case basis requests to"
+            + " remove the link to that site from its indices. However,"
+            + " if the operator of the site does not take steps to"
+            + " prevent it, the automatic facilities used to create"
+            + " the indices are likely to find that site and index it"
+            + " again in a relatively short amount of time.";
+    executeUpdate("create table data(colname clob)");
+    String sql = "insert into data(colname) values (?)";
+    try (PreparedStatement ps = getConnection().prepareStatement(sql)) {
+      ps.setBinaryStream(1, new ByteArrayInputStream(clobData.getBytes(UTF_8)));
```
wiarlawd: setString(1, clobData) is fine, but never setBinaryStream.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -428,4 +438,86 @@ public void testNullBLOB() throws Exception {
...
+        + "</table_rec>"
+        + "</table>"
+        + "</database>";
+    ResultSet rs = executeQueryAndNext("select * from data");
+    String result = generateXml(rs);
+    assertEquals(golden, result);
+  }
+
+  @Test
+  public void testmultipleTypes() throws Exception {
```
wiarlawd: Capitalize "M"

wiarlawd (changes requested):
One coding issue, the rest are names. Please changed added tests everywhere to match the style guide ("testBOOLEAN" is not allowed) and our test conventions ("testSomething_testCase" rather than "testTestCaseSomething").

src/com/google/enterprise/adaptor/database/TupleReader.java
```diff
@@ -202,48 +272,78 @@ protected void emitRow(ResultSet resultSet, ContentHandler handler)
...
+          }
           break;
-        case Types.CLOB:
         case Types.LONGVARCHAR:
-          Reader reader = resultSet.getCharacterStream(col);
-          handler.startElement("", columnName, columnName, atts);
-          copyValidXMLCharacters(reader, outWriter);
-          reader.close();
+          try (Reader reader = resultSet.getCharacterStream(col);
+              Writer writer =
```
wiarlawd: s/writer/outWriter/ for this case.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -287,4 +421,103 @@ public void testBLOB() throws Exception {
...
+        + "</table>"
+        + "</database>";
+    ResultSet rs = executeQueryAndNext("select * from data");
+    String result = generateXml(rs);
+    assertEquals(golden, result);
+  }
+
+  @Test
+  public void testMultipleTypes() throws Exception {
+    executeUpdate("create table data(id integer, name varchar, "
```
wiarlawd: Break before these two string to avoid string concatenation (prefer higher-level breaks).

wiarlawd (approved):
LGTM++ with two nits.

Please change commit title to "Use try-with-resources and canonical getters in TupleReader" and remove the description entirely. This is more specific about the changes.

PR #56, opened 2017-06-03
UniqueKey can accept types taken from ResultSetMetaData

BrettMichaelJohnson:
This is a first cut.

### DIFF
--- a/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
+++ b/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
@@ -273,12 +273,16 @@ public class DatabaseAdaptor extends AbstractAdaptor {
 
     // Verify all column names.
     try (Connection conn = makeNewConnection()) {
-      verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
-          "db.uniqueKey", uniqueKey.getDocIdSqlColumns());
-      verifyColumnNames(conn, "db.singleDocContentSql", singleDocContentSql,
-          "db.singleDocContentSqlParameters", uniqueKey.getContentSqlColumns());
-      verifyColumnNames(conn, "db.aclSql", aclSql,
-          "db.aclSqlParameters", uniqueKey.getAclSqlColumns());
+      uniqueKey.addColumnTypes(
+          verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
+              "db.uniqueKey", uniqueKey.getDocIdSqlColumns()));
+      uniqueKey.addColumnTypes(
+          verifyColumnNames(conn, "db.singleDocContentSql", singleDocContentSql,
+              "db.singleDocContentSqlParameters",
+              uniqueKey.getContentSqlColumns()));
+      uniqueKey.addColumnTypes(
+          verifyColumnNames(conn, "db.aclSql", aclSql,
+              "db.aclSqlParameters", uniqueKey.getAclSqlColumns()));
       if (!actionColumn.isEmpty()) {
         verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
             "db.actionColumn", Arrays.asList(actionColumn));
@@ -317,18 +321,21 @@ public class DatabaseAdaptor extends AbstractAdaptor {
    * @param sql the SQL query; the query is prepared, but not executed
    * @param columnConfig the configuration property for the column names
    * @param columnNames the column names to verify
+   * @return a Map of columnName to Integer java.sql.Types
    * @throws InvalidConfigurationException if any of the columns are not found
    */
   @VisibleForTesting
-  static void verifyColumnNames(Connection conn, String sqlConfig, String sql,
-      String columnConfig, Collection<String> columnNames) {
+  static Map<String, Integer> verifyColumnNames(Connection conn,
+       String sqlConfig, String sql, String columnConfig,
+       Collection<String> columnNames) {
+    Map<String, Integer> typeMap = new HashMap<>();
     if (isNullOrEmptyString(sql)) {
       log.log(Level.FINEST,
           "Skipping validation of empty query {0}", sqlConfig);
-      return;
+      return typeMap;
     }
     if (columnNames.isEmpty()) {
-      return;
+      return typeMap;
     }
     log.log(Level.FINEST, "Looking for columns {0} in {1}",
         new Object[] { columnNames, sqlConfig });
@@ -357,6 +364,8 @@ public class DatabaseAdaptor extends AbstractAdaptor {
               "Matched column \"{0}\" as \"{1}\" in query {2}",
               new Object[] { match, actual, sqlConfig });
           targets.remove(match);
+          // TODO(bmj): What happens with MySQL here.
+          typeMap.put(match, rsmd.getColumnType(i));
         }
       }
       if (!targets.isEmpty()) {
@@ -375,6 +384,7 @@ public class DatabaseAdaptor extends AbstractAdaptor {
           "Unable to validate configured column names for query {0}: {1}",
           new Object[] { sqlConfig, e });
     }
+    return typeMap;
   }
 
   /** Get all doc ids from database. */
--- a/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
+++ b/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
@@ -273,16 +273,15 @@ public class DatabaseAdaptor extends AbstractAdaptor {
 
     // Verify all column names.
     try (Connection conn = makeNewConnection()) {
-      uniqueKey.addColumnTypes(
+      Map<String, Integer> columnTypes =
           verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
-              "db.uniqueKey", uniqueKey.getDocIdSqlColumns()));
-      uniqueKey.addColumnTypes(
-          verifyColumnNames(conn, "db.singleDocContentSql", singleDocContentSql,
-              "db.singleDocContentSqlParameters",
-              uniqueKey.getContentSqlColumns()));
-      uniqueKey.addColumnTypes(
-          verifyColumnNames(conn, "db.aclSql", aclSql,
-              "db.aclSqlParameters", uniqueKey.getAclSqlColumns()));
+              "db.uniqueKey", uniqueKey.getDocIdSqlColumns());
+      uniqueKey.addColumnTypes(columnTypes);
+      verifyColumnNames(conn, "db.singleDocContentSql", singleDocContentSql,
+          "db.singleDocContentSqlParameters",
+          uniqueKey.getContentSqlColumns());
+      verifyColumnNames(conn, "db.aclSql", aclSql,
+          "db.aclSqlParameters", uniqueKey.getAclSqlColumns());
       if (!actionColumn.isEmpty()) {
         verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
             "db.actionColumn", Arrays.asList(actionColumn));
--- a/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
+++ b/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
@@ -278,8 +278,7 @@ public class DatabaseAdaptor extends AbstractAdaptor {
               "db.uniqueKey", uniqueKey.getDocIdSqlColumns());
       uniqueKey.addColumnTypes(columnTypes);
       verifyColumnNames(conn, "db.singleDocContentSql", singleDocContentSql,
-          "db.singleDocContentSqlParameters",
-          uniqueKey.getContentSqlColumns());
+          "db.singleDocContentSqlParameters", uniqueKey.getContentSqlColumns());
       verifyColumnNames(conn, "db.aclSql", aclSql,
           "db.aclSqlParameters", uniqueKey.getAclSqlColumns());
       if (!actionColumn.isEmpty()) {
@@ -363,7 +362,6 @@ public class DatabaseAdaptor extends AbstractAdaptor {
               "Matched column \"{0}\" as \"{1}\" in query {2}",
               new Object[] { match, actual, sqlConfig });
           targets.remove(match);
-          // TODO(bmj): What happens with MySQL here.
           typeMap.put(match, rsmd.getColumnType(i));
         }
       }
--- a/src/com/google/enterprise/adaptor/database/UniqueKey.java
+++ b/src/com/google/enterprise/adaptor/database/UniqueKey.java
@@ -208,7 +208,8 @@ class UniqueKey {
     }
     if (!badColumns.isEmpty()) {
       throw new InvalidConfigurationException("Unknown column type for the"
-          + " following columns: " + badColumns);
+          + " following columns: " + badColumns
+          + ". Please set explicit types in db.uniqueKey.");
     }
   }
 
@@ -243,13 +244,10 @@ class UniqueKey {
           part = "" + rs.getLong(name);
           break;
         default:
-          String errmsg = "Unknown column type for '" + name + "'. "
-              + "Please set an explicit column type in db.uniqueKey.";
-          throw new AssertionError(errmsg);
+          throw new AssertionError("Invalid type for column " + name
+              + ": " + types.get(name));
       }
-      if (encode) {
-        part = encodeSlashInData(part);
-      }
+      part = encodeSlashInData(part);
       sb.append("/").append(part);
     }
     return sb.toString().substring(1);
@@ -295,9 +293,8 @@ class UniqueKey {
           st.setLong(i + 1, Long.parseLong(valueOfCol));
           break;
         default:
-          String errmsg = "Unknown column type for '" + colName + "'. "
-              + "Please set an explicit column type in db.uniqueKey.";
-          throw new AssertionError(errmsg);
+          throw new AssertionError("Invalid type for column " + colName
+              + ": " + typeOfCol);
       }
     }
   }
--- a/src/com/google/enterprise/adaptor/database/UniqueKey.java
+++ b/src/com/google/enterprise/adaptor/database/UniqueKey.java
@@ -44,8 +44,7 @@ class UniqueKey {
     TIMESTAMP,
     DATE,
     TIME,
-    LONG,
-    UNKNOWN
+    LONG
   }
 
   private final List<String> names;  // columns used for DocId
@@ -79,18 +78,17 @@ class UniqueKey {
       String def[] = e.split(":", 2);
       String name = def[0].trim();
       ColumnType type;
-
       if (def.length == 1) {
         // No type given. Try to get type from ResultSetMetaData later.
-        type = ColumnType.UNKNOWN;
+        type = null;
       } else {
         String typeStr = def[1].trim();
         try {
           type = ColumnType.valueOf(typeStr.toUpperCase(US));
         } catch (IllegalArgumentException iae) {
-          String errmsg = "Invalid UniqueKey type '" + typeStr + "' for "
-              + "name. Valid types are: int, string, timestamp, date, time, "
-              + "and long.";
+          String errmsg = "Invalid UniqueKey type '" + typeStr
+              + "' for '" + name + "'. Valid types are: "
+              + ColumnType.values().toString().toLowerCase(US);
           throw new InvalidConfigurationException(errmsg);
         }
       }
@@ -98,11 +96,11 @@ class UniqueKey {
         String errmsg = "Invalid db.uniqueKey configuration: key name '"
             + name + "' was repeated.";
         throw new InvalidConfigurationException(errmsg);
-      } else if (tmpTypes.containsKey(name) && tmpTypes.get(name) != type) {
+      } else if (tmpTypes.containsKey(name)
+                 && (tmpTypes.get(name) != type || type == null)) {
         // The ukDecls contain two keys that differ only in case but are of
         // different types. Force a case-sensitive type lookup by replacing
         // the partial case-insensitive type map with a case-sensitive copy.
-        // TODO(bmj): What to do if they are both UNKNOWN?
         Map<String, ColumnType> caseSensitiveTypes = new TreeMap<>();
         for (String nombre : tmpNames) {
           caseSensitiveTypes.put(nombre, tmpTypes.get(nombre));
@@ -142,7 +140,7 @@ class UniqueKey {
     for (String name : cols.split(",", 0)) {
       name = name.trim();
       if (!validNames.contains(name)) {
-        String errmsg = "Unknown column '" + name + "' from "+ paramConfig
+        String errmsg = "Unknown column '" + name + "' from " + paramConfig
             + " not found in db.uniqueKey: " + validNames;
         throw new InvalidConfigurationException(errmsg);
       }
@@ -167,18 +165,20 @@ class UniqueKey {
    * Attempt to extract ColumnTypes from the supplied Map
    * of column names to java.sql.Type
    *
-   * @param typeMap Map of column names to Integer java.sql.Types
+   * @param sqlTypes Map of column names to Integer java.sql.Types
    */
-  void addColumnTypes(Map<String, Integer> typeMap) {
-    for (Map.Entry<String, Integer> entry : typeMap.entrySet()) {
-      if (types.get(entry.getKey()) != ColumnType.UNKNOWN) {
+  void addColumnTypes(Map<String, Integer> sqlTypes) {
+    for (Map.Entry<String, Integer> entry : sqlTypes.entrySet()) {
+      if (types.get(entry.getKey()) != null) {
         continue;
       }
       ColumnType type;
       switch (entry.getValue()) {
-        case Types.INTEGER:
-        case Types.SMALLINT:
+        case Types.BIT:
+        case Types.BOOLEAN:
         case Types.TINYINT:
+        case Types.SMALLINT:
+        case Types.INTEGER:
           type = ColumnType.INT;
           break;
         case Types.BIGINT:
@@ -187,6 +187,10 @@ class UniqueKey {
         case Types.CHAR:
         case Types.VARCHAR:
         case Types.LONGVARCHAR:
+        case Types.NCHAR:
+        case Types.NVARCHAR:
+        case Types.LONGNVARCHAR:
+        case Types.DATALINK:
           type = ColumnType.STRING;
           break;
         case Types.DATE:
@@ -199,9 +203,9 @@ class UniqueKey {
           type = ColumnType.TIMESTAMP;
           break;
         default:
-          String errmsg = "Invalid UniqueKey type '" + entry.getValue() + "'"
-              + " for " + entry.getKey() + ". Valid types are: int, string,"
-              + " timestamp, date, time and long.";
+          String errmsg = "Invalid UniqueKey SQLtype '" + entry.getValue() + "'"
+              + " for " + entry.getKey() + ". Please set an explicit key type"
+              + " in db.uniqueKey.";
           throw new InvalidConfigurationException(errmsg);
       }
       types.put(entry.getKey(), type);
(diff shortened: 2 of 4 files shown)


src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -273,12 +273,16 @@ public void init(AdaptorContext context) throws Exception {
 
     // Verify all column names.
     try (Connection conn = makeNewConnection()) {
-      verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
-          "db.uniqueKey", uniqueKey.getDocIdSqlColumns());
-      verifyColumnNames(conn, "db.singleDocContentSql", singleDocContentSql,
-          "db.singleDocContentSqlParameters", uniqueKey.getContentSqlColumns());
-      verifyColumnNames(conn, "db.aclSql", aclSql,
-          "db.aclSqlParameters", uniqueKey.getAclSqlColumns());
+      uniqueKey.addColumnTypes(
```
wiarlawd: I'm usually a fan of nesting calls, but not when the *primary effect* of the buried call is a side-effect, as in this case. I pondered a local Map variable, but I think I prefer passing uniqueKey (or null) to verifyColumnNames, and having verifyColumnNames push the map.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -357,6 +364,8 @@ static void verifyColumnNames(Connection conn, String sqlConfig, String sql,
               "Matched column \"{0}\" as \"{1}\" in query {2}",
               new Object[] { match, actual, sqlConfig });
           targets.remove(match);
+          // TODO(bmj): What happens with MySQL here.
+          typeMap.put(match, rsmd.getColumnType(i));
```
wiarlawd: I had gotten as far as realizing that we could do the work with ResultSetMetaData rather than ParameterMetaData, but hadn't realized that we could avoid having to figure out which column the type was for this way.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -357,6 +364,8 @@ static void verifyColumnNames(Connection conn, String sqlConfig, String sql,
               "Matched column \"{0}\" as \"{1}\" in query {2}",
               new Object[] { match, actual, sqlConfig });
           targets.remove(match);
+          // TODO(bmj): What happens with MySQL here.
```
wiarlawd: The MySQL issue is with ParameterMetaData, not ResultSetMetaData. Not to say that this couldn't blow up, and since it's optional we likely want a try-catch anyway.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -75,22 +77,22 @@
...
+      if (def.length == 1) {
+        // No type given. Try to get type from ResultSetMetaData later.
+        type = ColumnType.UNKNOWN;
+      } else {
+        String typeStr = def[1].trim();
+        try {
+          type = ColumnType.valueOf(typeStr.toUpperCase(US));
+        } catch (IllegalArgumentException iae) {
+          String errmsg = "Invalid UniqueKey type '" + typeStr + "' for "
+              + "name. Valid types are: int, string, timestamp, date, time, "
```
wiarlawd: I had thought to comment previously on the uppercase nature of ColumnType.values(), but didn't. I thought it was better than listing the names like this, and not worth lowercasing the values for the log message.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -75,22 +77,22 @@
...
-      String name = def[0];
+      String name = def[0].trim();
       ColumnType type;
-      try {
-        type = ColumnType.valueOf(def[1].toUpperCase(US));
-      } catch (IllegalArgumentException iae) {
-        String errmsg = "Invalid UniqueKey type '" + def[1] + "'. Valid"
-            + " types are: " + ColumnType.values();
-        throw new InvalidConfigurationException(errmsg);
+
```
wiarlawd: Strange blank line. I especially don't like blank lines between uninitialized variables and the code that initializes them.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -100,6 +102,7 @@
         // The ukDecls contain two keys that differ only in case but are of
         // different types. Force a case-sensitive type lookup by replacing
         // the partial case-insensitive type map with a case-sensitive copy.
+        // TODO(bmj): What to do if they are both UNKNOWN?
```
wiarlawd: What choice do we have? I think it has to be treated as case-sensitive.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -160,6 +163,51 @@
...
+        case Types.SMALLINT:
+        case Types.TINYINT:
+          type = ColumnType.INT;
+          break;
+        case Types.BIGINT:
+          type = ColumnType.LONG;
+          break;
+        case Types.CHAR:
+        case Types.VARCHAR:
+        case Types.LONGVARCHAR:
```
wiarlawd, line 179: I think we should include the NCHAR variations and DATALINK (an URL, and the spec says getString/setString are supported).

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -160,6 +163,51 @@
...
+          type = ColumnType.DATE;
+          break;
+        case Types.TIME:
+          type = ColumnType.TIME;
+          break;
+        case Types.TIMESTAMP:
+          type = ColumnType.TIMESTAMP;
+          break;
+        default:
+          String errmsg = "Invalid UniqueKey type '" + entry.getValue() + "'"
```
wiarlawd: I think s/type/SQL type/ and maybe replace the second sentence with something like: "Set an explicit key type in db.uniqueKey."

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -160,6 +163,51 @@
...
+              + " for " + entry.getKey() + ". Valid types are: int, string,"
+              + " timestamp, date, time and long.";
+          throw new InvalidConfigurationException(errmsg);
+      }
+      types.put(entry.getKey(), type);
+    }
+  }
+
   String makeUniqueId(ResultSet rs, boolean encode) throws SQLException {
```
wiarlawd: Future work to add checks for UNKNOWN to the methods accessing types after init?

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -160,6 +163,51 @@
...
+  void addColumnTypes(Map<String, Integer> typeMap) {
+    for (Map.Entry<String, Integer> entry : typeMap.entrySet()) {
+      if (types.get(entry.getKey()) != ColumnType.UNKNOWN) {
+        continue;
+      }
+      ColumnType type;
+      switch (entry.getValue()) {
+        case Types.INTEGER:
+        case Types.SMALLINT:
+        case Types.TINYINT:
```
wiarlawd: Convention sorts these small-to-large, like the character types are below. Also, I think we should add BIT and BOOLEAN here (getInt/setInt should work for them both).

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -160,6 +163,51 @@
     return aclSqlCols;
   }
 
+  /**
+   * Attempt to extract ColumnTypes from the supplied Map
+   * of column names to java.sql.Type
+   *
+   * @param typeMap Map of column names to Integer java.sql.Types
+   */
+  void addColumnTypes(Map<String, Integer> typeMap) {
```
wiarlawd: Maybe s/typeMap/sqlTypes/.

wiarlawd (commented):
Lots of comments, but I really like this.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -273,12 +273,16 @@ public void init(AdaptorContext context) throws Exception {
...
-      verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
-          "db.uniqueKey", uniqueKey.getDocIdSqlColumns());
-      verifyColumnNames(conn, "db.singleDocContentSql", singleDocContentSql,
-          "db.singleDocContentSqlParameters", uniqueKey.getContentSqlColumns());
-      verifyColumnNames(conn, "db.aclSql", aclSql,
-          "db.aclSqlParameters", uniqueKey.getAclSqlColumns());
+      uniqueKey.addColumnTypes(
+          verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
+              "db.uniqueKey", uniqueKey.getDocIdSqlColumns()));
+      uniqueKey.addColumnTypes(
```
wiarlawd: I think these next two are redundant. All of the columns have to appear in db.uniqueKey, and if the type map is case-sensitive, then all of the case-sensitive keys have to appear there as well.

src/com/google/enterprise/adaptor/database/UniqueKey.java
wiarlawd: But UNKNOWN is now an enum value, which we don't want to log as valid. OK, so that creates these logging problems, and it also creates a problem with the try block, because UNKNOWN should not be allowed in a typeStr. I had thought about suggesting null map values instead, but I did like the explicit UNKNOWN. If UNKNOWN causes at least two problems that null does not, though, maybe null is a winner.

src/com/google/enterprise/adaptor/database/UniqueKey.java
wiarlawd: I see the checks are already there, but maybe we should throw an InvalidConfigurationException instead of AssertionError now?

wiarlawd (commented):
Some followup to earlier comments.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -75,22 +77,22 @@
...
       ColumnType type;
-      try {
-        type = ColumnType.valueOf(def[1].toUpperCase(US));
-      } catch (IllegalArgumentException iae) {
-        String errmsg = "Invalid UniqueKey type '" + def[1] + "'. Valid"
-            + " types are: " + ColumnType.values();
-        throw new InvalidConfigurationException(errmsg);
+
+      if (def.length == 1) {
+        // No type given. Try to get type from ResultSetMetaData later.
```
PapaJoltOmega: have we any evidence that providing the types has been a challange?

PapaJoltOmega (commented):
Thank you

src/com/google/enterprise/adaptor/database/UniqueKey.java
wiarlawd: Only some confusion about uniqueKey vs metadataColumns. I've seen config files that had things like: db.metadataColumns=thiscolumn:int, thatcolumn:string.

More generally, people have lots of problems configuring the database connector, and we've been looking at the bug reports, support cases, and also conversations with QA and support. Beyond those specifics, we've been trying to make it both easier to configure in whatever ways we could (e.g., allowing the identity mapping by default in db.metadataColumns), and have the connector be more specific about what's wrong when there are config errors.

Back to the specifics, this is a case where the database knows the data we need, and we now even look at the data we need in init, so it seems a little silly not to use the data to simplify the configuration.

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
BrettMichaelJohnson: Cleaned up.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java
BrettMichaelJohnson: Converted from UNKNOWN to null, and revamped the error messages.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -239,9 +243,8 @@ String makeUniqueId(ResultSet rs, boolean encode) throws SQLException {
           part = "" + rs.getLong(name);
           break;
         default:
-          String errmsg = "Invalid type '" + types.get(name) + "' for '"
-              + name + "'. Valid types are: int, string, timestamp, date"
-              + ", time and long.";
+          String errmsg = "Unknown column type for '" + name + "'. "
```
wiarlawd: I think my two comments here didn't go together. With UNKNOWN, this was a config error, but now it's back to an AssertionError (it shouldn't happen, because every enum value is accounted for).

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -292,9 +295,8 @@ private void setSqlValues(PreparedStatement st, String uniqueId,
           st.setLong(i + 1, Long.parseLong(valueOfCol));
           break;
         default:
-          String errmsg = "Invalid column type: `" + typeOfCol + "' for '"
-              + colName + "'. Valid column types are int, string, timestamp,"
-              + "date, time and long.";
+          String errmsg = "Unknown column type for '" + colName + "'. "
```
wiarlawd: Same here.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -273,10 +273,13 @@ public void init(AdaptorContext context) throws Exception {
...
     try (Connection conn = makeNewConnection()) {
-      verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
-          "db.uniqueKey", uniqueKey.getDocIdSqlColumns());
+      Map<String, Integer> columnTypes =
+          verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
+              "db.uniqueKey", uniqueKey.getDocIdSqlColumns());
+      uniqueKey.addColumnTypes(columnTypes);
       verifyColumnNames(conn, "db.singleDocContentSql", singleDocContentSql,
-          "db.singleDocContentSqlParameters", uniqueKey.getContentSqlColumns());
+          "db.singleDocContentSqlParameters",
```
wiarlawd: Rejoin lines.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
wiarlawd: This TODO can go away. MySQL doesn't have a problem here. If some other database does, then either we can decide we want a separate try-catch here, to allow the unique key types to fail independently of the rest of the validation (which is already explicitly not failing by time we get here), or just ignore errors here and bail on the validation.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -357,6 +363,8 @@ static void verifyColumnNames(Connection conn, String sqlConfig, String sql,
               "Matched column \"{0}\" as \"{1}\" in query {2}",
               new Object[] { match, actual, sqlConfig });
           targets.remove(match);
+          // TODO(bmj): What happens with MySQL here.
+          typeMap.put(match, rsmd.getColumnType(i));
```
wiarlawd: This has a problem with case-sensitive UniqueKey.types maps. I think in UniqueKey, if the type map is case-sensitive then we just can't use this information. I was wrong that this approach solved the problem of matching the types to the unique key columns.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -75,28 +76,28 @@
...
+      String name = def[0].trim();
       ColumnType type;
-      try {
-        type = ColumnType.valueOf(def[1].toUpperCase(US));
-      } catch (IllegalArgumentException iae) {
-        String errmsg = "Invalid UniqueKey type '" + def[1] + "'. Valid"
-            + " types are: " + ColumnType.values();
-        throw new InvalidConfigurationException(errmsg);
+      if (def.length == 1) {
+        // No type given. Try to get type from ResultSetMetaData later.
```
PapaJoltOmega, line 82: why do want to figure it out?  why not continue to require that the types be provided?

PapaJoltOmega: I just now see that jlacey replied earlier.  The code seems to have changed and the comments (mine and jlacey replay) were hidden by github.  reading the reply now....

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -75,28 +76,28 @@
...
+        }
       }
       if (tmpNames.contains(name)) {
         String errmsg = "Invalid db.uniqueKey configuration: key name '"
             + name + "' was repeated.";
         throw new InvalidConfigurationException(errmsg);
-      } else if (tmpTypes.containsKey(name) && tmpTypes.get(name) != type) {
+      } else if (tmpTypes.containsKey(name)
+                 && (tmpTypes.get(name) != type || type == null)) {
         // The ukDecls contain two keys that differ only in case but are of
```
wiarlawd: s/are of/may be/

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -160,6 +161,57 @@
...
   }
 
+  /**
+   * Attempt to extract ColumnTypes from the supplied Map
+   * of column names to java.sql.Type
+   *
+   * @param sqlTypes Map of column names to Integer java.sql.Types
+   */
+  void addColumnTypes(Map<String, Integer> sqlTypes) {
+    for (Map.Entry<String, Integer> entry : sqlTypes.entrySet()) {
```
wiarlawd, line 161: We can get the Comparator of a TreeMap, and it's null if the natural ordering (aka case-sensitive) is used, so we don't need a separate flag to see if we can apply the SQL types here.

wiarlawd (changes requested):

src/com/google/enterprise/adaptor/database/UniqueKey.java
PapaJoltOmega: ok.

so processing of metadata and unique key becomes more uniform. that seems good. 

addressing specific problems you've seen customers have with config is great!

i guess the remaining part is to add these checks in as simple way as possible. thank you for keeping the code as simple to follow as possible.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -160,6 +161,57 @@
     return aclSqlCols;
   }
 
+  /**
+   * Attempt to extract ColumnTypes from the supplied Map
+   * of column names to java.sql.Type
+   *
+   * @param sqlTypes Map of column names to Integer java.sql.Types
+   */
+  void addColumnTypes(Map<String, Integer> sqlTypes) {
```
PapaJoltOmega, line 160: can this be something that is part of constructor?  

Adding this first mutator makes this class no longer immutable.  

Database adaptor could replace its instance of UniqueKey, rather than modify its unique key, maybe?

PapaJoltOmega (commented):
Thank you

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java
BrettMichaelJohnson: As discussed, we will make case-insensitive duplicate keys configuration error, and avoid all this.

src/com/google/enterprise/adaptor/database/UniqueKey.java
BrettMichaelJohnson: I will add a UniqueKey.Builder in a separate CL.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -160,6 +151,67 @@
...
+    }
+    ArrayList<String> badColumns = new ArrayList<>();
+    for (Map.Entry<String, ColumnType> entry : types.entrySet()) {
+      if (entry.getValue() == null) {
+        badColumns.add(entry.getKey());
+      }
+    }
+    if (!badColumns.isEmpty()) {
+      throw new InvalidConfigurationException("Unknown column type for the"
+          + " following columns: " + badColumns);
```
wiarlawd: How about either: "Missing column types for " + badColumns + ". Please set explicit types in db.uniqueKey." 
Or: "Missing column types for the following columns in db.uniqueKey: " + badColumns
I slightly prefer the former.

src/com/google/enterprise/adaptor/database/UniqueKey.java
wiarlawd: Sorry, missed a thing here, we should definitely log types.get(name) here. I think the original message (without the valid types) or similar is fine for an AssertionError:
"Invalid type '" + types.get(name) + "' for '" + name + "' in db.uniqueKey."

wiarlawd (commented):
Wordsmithing on error messages (in particular, to include db.uniqueKey in user error messages, and the failed types in the assertion errors).

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -317,18 +319,21 @@ public void init(AdaptorContext context) throws Exception {
    * @param sql the SQL query; the query is prepared, but not executed
    * @param columnConfig the configuration property for the column names
    * @param columnNames the column names to verify
+   * @return a Map of columnName to Integer java.sql.Types
```
PapaJoltOmega, line 325: can that be Map<String, UniqueKey.ColumnType>?  It does seem a bit unfortunate that SQL types are Integer. What do you think?

src/com/google/enterprise/adaptor/database/UniqueKey.java
PapaJoltOmega: Will the type map be the only optional builder parameter? Maybe additional ctor is enough? If there will be more optional parameters then maybe builder is the right idea.

PapaJoltOmega (commented):
Cool.  Thank you for improving the configuration experience.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
BrettMichaelJohnson: Look at how UniqueKey maps multiple SQL types to ColumnType.  Moving that logic here seems very out-of-place.  But that does not really matter.  When I make a UniqueKey.Builder, it will likely take a ResultSet (or ResultSetMetaData), rather than this map.

src/com/google/enterprise/adaptor/database/UniqueKey.java
BrettMichaelJohnson: At this point, I think you get here when DB connectivity is down, so I was hesitant to use the second sentence. But the more I think about it. No connectivity would result in a SQLException, so you wouldn't get here.

src/com/google/enterprise/adaptor/database/UniqueKey.java
BrettMichaelJohnson: types.get() returns null here. That is why I changed the message when going from UNKNOWN to null.

PapaJoltOmega:
>
>
>  Look at how UniqueKey maps multiple SQL types to ColumnType. Moving that
> logic here seems very out-of-place. But that does not really matter. When I
> make a UniqueKey.Builder, it will likely take a ResultSet (or
> ResultSetMetaData), rather than this map.
>

​SGTM.  Thanks.​

src/com/google/enterprise/adaptor/database/UniqueKey.java
wiarlawd: There are other reasons to want the Builder. There are number of methods that are used either only from init, while the unique key is being built, or from the tests, poking at the details of the parsing. We can put all of those on the Builder class and reduce the actual UniqueKey to its post-construction production methods. (We can also add some tests of the production methods using crafted prepared statements.) Also, we are thinking of splitting the db.singleDocContentSqlParameters and db.aclSqlParameters arguments into separate setters on the builder, because they like everything else rely on the parsed db.uniqueKey, which will be the sole builder constructor argument.

PapaJoltOmega:
​I see the vision now.  Sounds good.  Thank you.​

-
- technology's compounding interest

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
wiarlawd: You would have to go back to the option you rejected to put the call to addColumnTypes inside verifyColumnNames. The result set is closed when the method exits, and while I think it would be unusual for the ResultSetMetaData to be tied to the ResultSet, I don't think there's anything in the JDBC spec that prevents that.

src/com/google/enterprise/adaptor/database/UniqueKey.java
wiarlawd: I mostly wanted to make sure that "db.uniqueKey" was mentioned in the message.

src/com/google/enterprise/adaptor/database/UniqueKey.java
wiarlawd: A null argument to switch throws a NPE. I actually think omitting the default on an enum switch is reasonable (and it's permitted by the style guide).

wiarlawd (approved):
LGTM++ with more argumentation (of a lesser nature) about the two log statements. ;-)

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java
BrettMichaelJohnson: Done. Then Undone, because compiler now complains that 'part' may be unassigned.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -191,14 +244,10 @@ String makeUniqueId(ResultSet rs, boolean encode) throws SQLException {
...
-              + name + "'. Valid types are: int, string, timestamp, date"
-              + ", time and long.";
-          throw new AssertionError(errmsg);
-      }
-      if (encode) {
-        part = encodeSlashInData(part);
+          throw new AssertionError("Invalid type for column " + name
+              + ": " + types.get(name));
       }
+      part = encodeSlashInData(part);
```
wiarlawd: Could inline this on the next line.

wiarlawd (approved):
LGTM++

PR #57, opened 2017-06-04
Improve code coverage for DatabaseAdaptor (part 1).

wiarlawd:
Part 1: Everything but ACLs, getModifiedDocIds, and authZ.
Improves coverage of DatabaseAdaptor from 69% to 75%.

Production changes:
* Expand logging ignored properties outside of lister mode.

Copy Logging.java from the Documentum connector to test log messages.

### DIFF
--- a/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
+++ b/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
@@ -150,6 +150,8 @@ public class DatabaseAdaptor extends AbstractAdaptor {
 
   @Override
   public void init(AdaptorContext context) throws Exception {
+    TreeSet<String> ignored = new TreeSet<>(); // Log ignored properties.
+
     Config cfg = context.getConfig();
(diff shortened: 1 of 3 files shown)


BrettMichaelJohnson (approved):

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -203,6 +205,9 @@ public void init(AdaptorContext context) throws Exception {
     } else {
       metadataColumns = new MetadataColumns(metadataColumnsConfig);
       log.config("metadata columns: " + metadataColumns);
+      if (includeAllColumnsAsMetadata) {
+        ignored.add("db.includeAllColumnsAsMetadata");
```
PapaJoltOmega, line 209: so specific columns overrides config to include all columns, yes?

wiarlawd: Yes. Seems odd to me, but it's the existing behavior. The mere existence of both flags seems odd to me (essentially duplicating the SQL select list in functionality).

PR #58, opened 2017-06-04
Improve code coverage for ValidatedUri.

wiarlawd:
The check of URI.isAbsolute leads to dead code. The URL constructor
with one argument does not allow relative URIs.

### DIFF
--- a/src/com/google/enterprise/adaptor/database/ValidatedUri.java
+++ b/src/com/google/enterprise/adaptor/database/ValidatedUri.java
@@ -42,7 +42,7 @@ class ValidatedUri {
     }
     try {
       // Basic syntax checking, with more understandable error messages.
-      // Also ensures the URI is a URL, not a URN.
+      // Also ensures the URI is a URL, not a URN, and is absolute.
       new URL(uriString);
       // Advanced syntax checking, with more cryptic error messages.
       uri = new URI(uriString);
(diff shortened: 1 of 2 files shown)


PapaJoltOmega (approved):

src/com/google/enterprise/adaptor/database/ValidatedUri.java
```diff
@@ -54,10 +54,6 @@ public ValidatedUri(String uriString) throws URISyntaxException {
       throw new URISyntaxException(uriString, reason);
     }
 
-    if (!uri.isAbsolute()) {
```
PapaJoltOmega, line 57: seeing how we missed that URL ctor detects this condition should we annotate ctor with someting like "// prevents relative URL"

test/com/google/enterprise/adaptor/database/ValidatedUriTest.java
```diff
@@ -59,15 +67,17 @@ public void testBadProtocol() throws Exception {
...
     new ValidatedUri("http://");
   }
 
   @Test
-  public void testRelativeUri() throws Exception {
+  public void testNoHost() throws Exception {
     thrown.expect(URISyntaxException.class);
-    new ValidatedUri("foo/bar");
+    thrown.expectMessage("no host");
+    new ValidatedUri("http:/foo/bar");
```
PapaJoltOmega: that looks weird


should that be http:///foo/bar?

BrettMichaelJohnson: It may look weird, but that is exactly what File.toURI() returns (except file: instead of http:). I do think we should try the /// variant, tho.

BrettMichaelJohnson (approved):

PR #65, opened 2017-06-09
Add a Builder for UniqueKey

BrettMichaelJohnson:
This moved the various getters used by verifyColumnNames
and the tests from UniqueKey to its Builder, leaving UniqueKey
instance with its basic functionality.

This moved UniqueKey creation to the bottom of the adaptor init()
method, after verifyColumnNames gets a chance to populate the
types map.

Other changes:
  - had to fix up some DatabaseAdaptor tests, since
    configuration validation happens in a different order.
  - added UniqueKey tests to bring code coverage up to 98%.
  - removed the various test getters, and replaced their usage
    with the regular getters.

### DIFF
--- a/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
+++ b/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
@@ -172,10 +172,6 @@ public class DatabaseAdaptor extends AbstractAdaptor {
     password = context.getSensitiveValueDecoder().decodeValue(
         cfg.getValue("db.password"));
 
-    boolean leaveIdAlone = new Boolean(cfg.getValue("docId.isUrl"));
-    encodeDocId = !leaveIdAlone;
-    log.config("encodeDocId: " + encodeDocId);
-
     everyDocIdSql = cfg.getValue("db.everyDocIdSql");
     log.config("every doc id sql: " + everyDocIdSql);
 
@@ -203,6 +199,12 @@ public class DatabaseAdaptor extends AbstractAdaptor {
     }
 
     modeOfOperation = cfg.getValue("db.modeOfOperation");
+    log.config("mode of operation: " + modeOfOperation);
+
+    boolean leaveIdAlone = new Boolean(cfg.getValue("docId.isUrl"));
+    encodeDocId = !leaveIdAlone;
+    log.config("encodeDocId: " + encodeDocId);
+
     if (modeOfOperation.equals("urlAndMetadataLister") && encodeDocId) {
       String errmsg = "db.modeOfOperation of \"" + modeOfOperation
           + "\" requires docId.isUrl to be \"true\"";
@@ -405,7 +407,7 @@ public class DatabaseAdaptor extends AbstractAdaptor {
         BufferedPusher outstream = new BufferedPusher(pusher)) {
       log.finer("queried for stream");
       while (rs.next()) {
-        DocId id = new DocId(uniqueKey.makeUniqueId(rs, encodeDocId));
+        DocId id = new DocId(uniqueKey.makeUniqueId(rs));
         DocIdPusher.Record.Builder builder = new DocIdPusher.Record.Builder(id);
         if (isDeleteAction(rs)) {
           builder.setDeleteFromIndex(true);
@@ -916,7 +918,7 @@ public class DatabaseAdaptor extends AbstractAdaptor {
             hasColumn(rs.getMetaData(), GsaSpecialColumns.GSA_TIMESTAMP);
         log.log(Level.FINEST, "hasTimestamp: {0}", hasTimestamp);
         while (rs.next()) {
-          DocId id = new DocId(uniqueKey.makeUniqueId(rs, encodeDocId));
+          DocId id = new DocId(uniqueKey.makeUniqueId(rs));
           DocIdPusher.Record.Builder builder =
               new DocIdPusher.Record.Builder(id).setCrawlImmediately(true);
           if ("urlAndMetadataLister".equals(modeOfOperation)) {
--- a/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
+++ b/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
@@ -176,14 +176,6 @@ public class DatabaseAdaptor extends AbstractAdaptor {
     encodeDocId = !leaveIdAlone;
     log.config("encodeDocId: " + encodeDocId);
 
-    uniqueKey = new UniqueKey(
-        cfg.getValue("db.uniqueKey"),
-        cfg.getValue("db.singleDocContentSqlParameters"),
-        cfg.getValue("db.aclSqlParameters"),
-        encodeDocId
-    );
-    log.config("primary key: " + uniqueKey);
-
     everyDocIdSql = cfg.getValue("db.everyDocIdSql");
     log.config("every doc id sql: " + everyDocIdSql);
 
@@ -277,18 +269,24 @@ public class DatabaseAdaptor extends AbstractAdaptor {
           "The following properties are set but will be ignored: " + ignored);
     }
 
+    UniqueKey.Builder ukBuilder
+        = new UniqueKey.Builder(cfg.getValue("db.uniqueKey"))
+        .setEncodeDocIds(encodeDocId)
+        .setContentSqlColumns(cfg.getValue("db.singleDocContentSqlParameters"))
+        .setAclSqlColumns(cfg.getValue("db.aclSqlParameters"));
+
     // Verify all column names.
     try (Connection conn = makeNewConnection()) {
       Map<String, Integer> columnTypes =
           verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
-              "db.uniqueKey", uniqueKey.getDocIdSqlColumns());
-      uniqueKey.addColumnTypes(columnTypes);
+              "db.uniqueKey", ukBuilder.getDocIdSqlColumns());
+      ukBuilder.addColumnTypes(columnTypes);
       verifyColumnNames(conn, "db.updateSql", updateSql,
-          "db.uniqueKey", uniqueKey.getDocIdSqlColumns());
+          "db.uniqueKey", ukBuilder.getDocIdSqlColumns());
       verifyColumnNames(conn, "db.singleDocContentSql", singleDocContentSql,
-          "db.singleDocContentSqlParameters", uniqueKey.getContentSqlColumns());
+          "db.singleDocContentSqlParameters", ukBuilder.getContentSqlColumns());
       verifyColumnNames(conn, "db.aclSql", aclSql,
-          "db.aclSqlParameters", uniqueKey.getAclSqlColumns());
+          "db.aclSqlParameters", ukBuilder.getAclSqlColumns());
       if (!actionColumn.isEmpty()) {
         verifyColumnNames(conn, "db.everyDocIdSql", everyDocIdSql,
             "db.actionColumn", Arrays.asList(actionColumn));
@@ -314,6 +312,9 @@ public class DatabaseAdaptor extends AbstractAdaptor {
     } catch (SQLException e) {
       log.log(Level.WARNING, "Unable to validate configured column names");
     }
+
+    uniqueKey = ukBuilder.build();
+    log.config("primary key: " + uniqueKey);
   }
 
   /**
--- a/src/com/google/enterprise/adaptor/database/UniqueKey.java
+++ b/src/com/google/enterprise/adaptor/database/UniqueKey.java
@@ -51,21 +51,22 @@ class UniqueKey {
   private final Map<String, ColumnType> types;  // types of DocId columns
   private final List<String> contentSqlCols;  // columns for content query
   private final List<String> aclSqlCols;  // columns for acl query
+  private final boolean encodeDocIds;
 
   private UniqueKey(List<String> docIdSqlCols, Map<String, ColumnType> types,
-      List<String> contentSqlCols, List<String> aclSqlCols) {
+      List<String> contentSqlCols, List<String> aclSqlCols,
+      boolean encodeDocIds) {
     this.docIdSqlCols = docIdSqlCols;
     this.types = types;
     this.contentSqlCols = contentSqlCols;
     this.aclSqlCols = aclSqlCols;
+    this.encodeDocIds = encodeDocIds;
   }
 
-  String makeUniqueId(ResultSet rs, boolean encode) throws SQLException {
-    if (!encode) {
-      if (docIdSqlCols.size() == 1) {
-        return rs.getString(docIdSqlCols.get(0));
-      }
-      throw new AssertionError("not encoding implies exactly one parameter");
+  String makeUniqueId(ResultSet rs) throws SQLException {
+    if (!encodeDocIds) {
+      // DocId must be a single string column.
+      return rs.getString(docIdSqlCols.get(0));
     }
     StringBuilder sb = new StringBuilder();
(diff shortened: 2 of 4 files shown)


src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -397,6 +220,247 @@ public boolean equals(Object other) {
 
   @Override
   public int hashCode() {
-    return java.util.Objects.hash(names, types, contentSqlCols, aclSqlCols);
+    return java.util.Objects.hash(docIdSqlCols, types, contentSqlCols, aclSqlCols);
+  }
+
+  /** Builder to create instances of {@code UniqueKey}. */
+  static class Builder {
+    private List<String> docIdSqlCols;  // columns used for DocId
```
wiarlawd: This can be final if you restore the tmpNames list.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -397,6 +220,247 @@ public boolean equals(Object other) {
...
+            type = ColumnType.valueOf(typeStr.toUpperCase(US));
+          } catch (IllegalArgumentException iae) {
+            String errmsg = "Invalid UniqueKey type '" + typeStr
+                + "' for '" + name + "'. Valid types are: "
+                + Arrays.toString(ColumnType.values()).toLowerCase(US);
+            throw new InvalidConfigurationException(errmsg);
+          }
+        }
+        docIdSqlCols.add(name);
+        if (types.put(name, type) != null) {
```
wiarlawd: Nice

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -397,6 +220,247 @@ public boolean equals(Object other) {
...
+          String errmsg = "Invalid db.uniqueKey configuration: key name '"
+              + name + "' was repeated.";
+          throw new InvalidConfigurationException(errmsg);
+        }
+      }
+      docIdSqlCols = Collections.unmodifiableList(docIdSqlCols);
```
wiarlawd: I think we should restore types as an unmodifiableMap as well. We can swizzle the whole map in addColumnTypes, but that way we never expose an editable map with the new getColumnTypes method.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -397,6 +220,247 @@ public boolean equals(Object other) {
...
+        if (entry.getValue() == null) {
+          badColumns.add(entry.getKey());
+        }
+      }
+      if (!badColumns.isEmpty()) {
+        throw new InvalidConfigurationException("Unknown column type for the"
+            + " following columns: " + badColumns
+            + ". Please set explicit types in db.uniqueKey.");
+      }
+      types = Collections.unmodifiableMap(types);
```
wiarlawd, line 439: As mentioned elsewhere, let's do this at the end of addColumnTypes.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -176,14 +176,6 @@ public void init(AdaptorContext context) throws Exception {
     encodeDocId = !leaveIdAlone;
     log.config("encodeDocId: " + encodeDocId);
 
-    uniqueKey = new UniqueKey(
```
wiarlawd: docId.isUrl moved here for the unique key. Please put it back down below modeOfOperation, just before leaveIdAlone is first used.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -47,121 +56,214 @@ public void dropAllObjects() throws SQLException {
...
+    new UniqueKey.Builder("num:int").setAclSqlColumns(null);
   }
 
   @Test
   public void testSingleInt() {
-    UniqueKey uk = newUniqueKey("numnum:int");
-    assertEquals(1, uk.numElementsForTest());
-    assertEquals("numnum", uk.nameForTest(0));
-    assertEquals(UniqueKey.ColumnType.INT, uk.typeForTest(0));
+    UniqueKey.Builder ukb = new UniqueKey.Builder("numnum:int");
```
wiarlawd: Personally, s/ukb/builder/. The use of "uk" was marginal for me (I think "key" would have been better, and I thought about the UK first about half the time I saw it), and the "ukb" is over the line.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -47,121 +56,214 @@ public void dropAllObjects() throws SQLException {
...
+    ukb.addColumnTypes(sqlTypes);
+  }
+
+  @Test
+  public void testEquals() throws Exception {
+    UniqueKey uk1 = new UniqueKey.Builder("id:int").build();
+    assertTrue(uk1.equals(new UniqueKey.Builder("id:int").build()));
+    assertFalse(uk1.equals(new UniqueKey.Builder("foo:int").build()));
+    assertFalse(uk1.equals(new UniqueKey.Builder("id:string").build()));
+    UniqueKey uk2 = new UniqueKey.Builder("id:int, other:string").build();
```
wiarlawd: Two other cases that strike me: that order in the names matters, and that calling the set*Columns methods with "" do not affect equality. Same for hashCode.

However, to play devil's advocate, why implement equals and hashCode at all? This is a class that has a single instance. Removing them avoids the need for these tests and 20 lines of prod code. Or, follow Effective Java Item 8 and implement both methods but throw an AssertionError to make sure they are never used. The testSpaces methods here would need to be changed to use the get*Columns methods, which seems more obvious anyway.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -178,6 +280,51 @@ public void testProcessingDocId() throws SQLException {
...
 
   @Test
+  public void testProcessingDocId_encodingError() throws SQLException {
+    executeUpdate("create table data(numnum int, strstr varchar)");
+    executeUpdate("insert into data(numnum, strstr) values(345, 'abc')");
+
+    try (Statement stmt = getConnection().createStatement();
+        ResultSet rs = stmt.executeQuery("select * from data")) {
+      UniqueKey uk = newUniqueKey("numnum:int,strstr:string");
+      assertTrue("ResultSet is empty", rs.next());
```
wiarlawd: We might as well use JdbcFixture.executeQueryAndNext in new code, at least, and save a few lines here (it encapsulates all the JDBC stuff from the try-with-resources to the rs.next call).

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -178,6 +280,51 @@ public void testProcessingDocId() throws SQLException {
...
+      thrown.handleAssertionErrors();
+      thrown.expect(AssertionError.class);
+      thrown.expectMessage("not encoding implies exactly one parameter");
+      uk.makeUniqueId(rs, /*encode=*/ false);
```
wiarlawd: Couldn't we just remove the encode parameter here and pass encodeDocIds from the builder? This AssertionError is dead code now that we use it to construct the UniqueKey.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -178,6 +280,51 @@ public void testProcessingDocId() throws SQLException {
...
+  }
+
+  @Test
+  public void testProcessingDocId_allTypes() throws SQLException {
+    executeUpdate("create table data(c1 integer, c2 bigint, c3 varchar, "
+        + "c4 date, c5 time, c6 timestamp)");
+    String sql = "insert into data(c1, c2, c3, c4, c5, c6) "
+        + "values (?, ?, ?, ?, ?, ?)";
+    Calendar gmt = Calendar.getInstance(TimeZone.getTimeZone("GMT"), US);
+    try (PreparedStatement ps = getConnection().prepareStatement(sql)) {
```
wiarlawd: Why do any of this with a PreparedStatement, and why use zeroes for the data-time values?

    executeUpdate("insert into data(c1, c2, c3, c4, c5, c6) "
        + "values (123, 4567890, 'foo', "
        + "{d '2007-08-09'}, {t '12:34:56'}, {ts '2007-08-09 12:34:56')");

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -178,6 +280,51 @@ public void testProcessingDocId() throws SQLException {
...
+      ps.setInt(1, 123);
+      ps.setLong(2, 4567890L);
+      ps.setString(3, "foo");
+      ps.setDate(4, new Date(0L), gmt);
+      ps.setTime(5, new Time(0L), gmt);
+      // TODO(bmj): setTimestamp with Calendar broken in H2?
```
wiarlawd: I love time zones. This isn't an H2 bug, it's just a behavior that some databases exhibit. Basically, the database isn't storing the time zone with the time and timestamp values. On the way in, it takes the given millis relative either to the given calendar, or to the local time zone. On the way out it does the same, but we're not specifying the time zone, so the local one is always used.

new Timestamp(0L), gmt => "1970-01-01 00:00:00"
"1970-01-01 00:00:00" => Timestamp.valueOf("1970-01-01 00:00:00"), local = "1970-01-01 00:00:00" = 28800000

new Timestamp(0L), local => "1969-12-31 16:00:00"
"1969-12-31 16:00:00" => Timestamp.valueOf("1969-12-31 16:00:00"), local = "1969-12-31 16:00:00" = 0

The problem isn't a bug with 0L and GMT, but with the fact that H2 doesn't store time zone information, so when turning "1970-01-01 00:00:00" back into a Timestamp, it uses the local time zone, and says that's 28800000 milliseconds UTC.

We should probably expand the role of db.updateTimestampTimezone, but in the meantime, we have three options:
1. Keep the test as-is, using local for timestamps because we get it back as milliseconds.
2. Use GMT for timestamps, too, but expect Timestamp.valueOf("1970-01-01").getTime() instead of 0 for the timestamp value in the docid.
3. Switch from new X(millis) to X.from(string) or use the JDBC escapes from the previous comment, and expect Timestamp.valueOf("1970-01-01").getTime() instead of 0 for the timestamp value in the docid.

Note that changing the prod code to use GMT isn't a solution, because we have no idea what time zone the actual dates in the database are in. Come to think of it, we should make sure that our timestamp values actually round-trip OK. If that's broken, we might be better off using the timestamp strings. I don't know why time uses them but timestamp does not. Or we might have to expand the use of db.updateTimestampTimezone sooner rather than later.

wiarlawd (commented):
Holy unnecessary diffs, Batman! Undoing just the moved code and renamed variables shrunk the diffs from 989 lines to 666. That's also a lot of history stomping without a lot of payoff. Why not just leave the Builder at the top of UniqueKey where the methods were?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
BrettMichaelJohnson: Done.

src/com/google/enterprise/adaptor/database/UniqueKey.java

src/com/google/enterprise/adaptor/database/UniqueKey.java
BrettMichaelJohnson: Thanks.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
BrettMichaelJohnson: I choose to nuke the methods and their tests.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -102,18 +100,18 @@ public void testEmptyThrows() {
...
+    assertEquals(ColumnType.INT, builder.getColumnTypes().get("other"));
   }
 
   @Test
   public void testIntString() {
-    UniqueKey.Builder ukb = new UniqueKey.Builder("numnum:int,strstr:string");
-    assertEquals(asList("numnum", "strstr"), ukb.getDocIdSqlColumns());
-    assertEquals(ColumnType.INT, ukb.getColumnTypes().get("numnum"));
-    assertEquals(ColumnType.STRING, ukb.getColumnTypes().get("strstr"));
+    UniqueKey.Builder builder = new UniqueKey.Builder("numnum:int,strstr:string");
```
wiarlawd: 80 cols

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -275,53 +241,23 @@ public void testProcessingDocId() throws SQLException {
...
         "c1:int, c2:long, c3:string, c4:date, c5:time, c6:timestamp").build();
-
-    try (Statement stmt = getConnection().createStatement();
-        ResultSet rs = stmt.executeQuery("select * from data")) {
-      assertTrue("ResultSet is empty", rs.next());
-      assertEquals("123/4567890/foo/1970-01-01/00:00:00/0",
-          uk.makeUniqueId(rs, /*encode=*/ true));
-    }
+    ResultSet rs = executeQueryAndNext("select * from data");
+    assertEquals("123/4567890/foo/2007-08-09/12:34:56/1186688096000",
```
wiarlawd: This is broken, because the milliseconds depend on your local time zone. That's why I suggested using Timestamp.valueOf("2007-08-09 12:34:56").getTime()  for the expected timestamp value.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -275,53 +241,23 @@ public void testProcessingDocId() throws SQLException {
...
-      ps.setString(3, "foo");
-      ps.setDate(4, new Date(0L), gmt);
-      ps.setTime(5, new Time(0L), gmt);
-      // TODO(bmj): setTimestamp with Calendar broken in H2?
-      ps.setTimestamp(6, new Timestamp(0L)/*, gmt*/);
-      assertEquals(1, ps.executeUpdate());
-    }
+    executeUpdate("insert into data(c1, c2, c3, c4, c5, c6) "
+        + "values (123, 4567890, 'foo', "
+        + "{d '2007-08-09'}, {t '12:34:56'}, {ts '2007-08-09T12:34:56'})");
```
wiarlawd: The missing "T" wasn't a typo. The JDBC escape syntax doesn't include it (it's optional in ISO 8601). No reason to depend on known H2-specific syntax unless we need it.

wiarlawd (changes requested):

wiarlawd (approved):
LGTM, but please wait until Monday to see if PJ wants to weigh in.

PR #74, opened 2017-06-13
Reorder SQL type switch statements in canonical order.

wiarlawd:
Order to match the tables in the JDBC API Tutorial and Reference 3/e:
numeric, boolean, character, binary, data-time, LOBs, and structured.
Within each group, smaller to larger. Date-time order is DATE, TIME,
TIMESTAMP. LOB order is CLOB, NCLOB, BLOB, SQLXML. Structured order is
ARRAY, REF, STRUCT, JAVA_OBJECT.

Unsupported types are grouped together in the last block before the
default label, in canonical order.

Reorder the tests to match the switch statements.

The only non-reordering change is to add BFILE to the switch in
TupleReader, matching the other three switches (RowToHtml,
ContentColumn, and DatabaseAdaptor).

### DIFF
--- a/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
+++ b/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
@@ -1832,113 +1832,6 @@ public class DatabaseAdaptorTest {
   }
 
   @Test
-  public void testMetadataColumns_clob() throws Exception {
-    // NCLOB shows up as CLOB in H2
-    String content = "Hello World";
-    executeUpdate("create table data(id int, content clob)");
-    String sql = "insert into data(id, content) values (1, ?)";
-    try (PreparedStatement ps = getConnection().prepareStatement(sql)) {
-      ps.setString(1, content);
-      assertEquals(1, ps.executeUpdate());
-    }
-
-    Map<String, String> configEntries = new HashMap<String, String>();
-    configEntries.put("db.uniqueKey", "ID:int");
-    configEntries.put("db.everyDocIdSql", "select * from data");
-    configEntries.put("db.singleDocContentSql",
-        "select * from data where id = ?");
-    configEntries.put("db.modeOfOperation", "rowToText");
-    configEntries.put("db.metadataColumns", "ID:col1, CONTENT:col2");
-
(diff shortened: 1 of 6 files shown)


test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2081,6 +1974,113 @@ public void testMetadataColumns_timestampNull() throws Exception {
...
+    RecordingResponse response = new RecordingResponse();
+    adaptor.getDocContent(request, response);
+
+    Metadata expected = new Metadata();
+    expected.add("col1", "1");
+    assertEquals(expected, response.getMetadata());
+  }
+
+  @Test
+  public void testMetadataColumns_integer() throws Exception {
```
BrettMichaelJohnson: Don't these two want to be up above testMetadataColumns_date()?

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -129,34 +129,48 @@ public void testVarchar() throws Exception {
...
-    assertEquals(
-        String.format(template, xml.replace("<", "&lt;").replace(">", "&gt;")),
-        result);
+    assertEquals(golden, result);
   }
 
   @Test
-  public void testVarchar_null() throws Exception {
-    executeUpdate("create table data(colname varchar)");
+  public void testBoolean() throws Exception {
```
BrettMichaelJohnson, line 149: Does this want to be above testInteger?

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
wiarlawd: There is no integer case in the addMetadata switch statement, so these are at the bottom to match the default case. I see they aren't at the bottom, so they should move, only down not up.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
wiarlawd: Below is right, but I was of a different mind when I didn't change the prod code. I had suggested putting these first in prod, but the canonical order has them in a second group (really third, after integers and reals). I'll change prod.

BrettMichaelJohnson (approved):

PR #80, opened 2017-07-11
Enhanced tests for optional types specified for UniqueKey

BrettMichaelJohnson:
In b/62535928 John suggests an additional test in DatabaseAdaptorTest that specifies a uniqueKey with multiple columns, some with types specified and some without.


PapaJoltOmega (approved):
Thank you

PR #81, opened 2017-07-12
Additional UniqueKey case-sensitivity tests and fix bug found

BrettMichaelJohnson:
Also fix a bug in UniqueKey.setSqlValues() found by the new tests.


PapaJoltOmega (approved):
Thank you

PR #84, opened 2017-07-14
Test aclSql using column aliases

BrettMichaelJohnson:


test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2374,6 +2392,43 @@ public void testGetDocContentAcl() throws Exception {
   }
 
   @Test
+  public void testGetDocContentAcl_columnAliases() throws Exception {
```
wiarlawd: Replace testGetDocContentAcl with this one.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -1201,18 +1201,36 @@ public void testInitVerifyColumnNames_aclSql() throws Exception {
   }
 
   @Test
+  public void testInitVerifyColumnNames_aclSql_columnAlias() throws Exception {
```
wiarlawd: Replace this with one testing aliases in metadataColumns.

wiarlawd (commented):
As discussed, let's take out the two bogus calls to verifyColumnNames that make sure that the parameters appear in the select lists for singleDocContentSql and aclSql, which is not actually required.

Add a test for getDocIds using a column alias for the unique key, and a test of getDocContent where the unique doesn't appear at all in the select list, as itself or the alias.

Obviously the commit messages needs to be generalized.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: done.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: done.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2345,20 +2363,57 @@ public void testMetadataColumns_varchar() throws Exception {
...
+        + " allowed_users varchar)");
+    executeUpdate("insert into acl(id, allowed_groups, allowed_users) "
         + "values (1001, 'pgroup1', 'puser1'), (1002, 'pgroup2', 'puser2')");
 
+    String aclSql = "select id, allowed_groups as gsa_permit_groups, "
+        + "allowed_users as gsa_permit_users from acl where id = ?";
+
+    ResultSet rs = executeQuery(aclSql.replace("?", "'1001'"));
+    ResultSetMetaData rsmd = rs.getMetaData();
+    assertEquals("allowed_groups", rsmd.getColumnName(2).toLowerCase());
```
wiarlawd: Use US locale.

wiarlawd (commented):
What about the other changes?

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: OK. So I accidentally made the validateColumnNames changes in the other CL.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -1150,7 +1152,9 @@ public void testInitVerifyColumnNames_uniqueKey_updateSql() throws Exception {
   }
 
   @Test
-  public void testInitVerifyColumnNames_singleDocContentSql() throws Exception {
+  public void
+      testInitVerifyColumnNames_singleDocContentSql_uniqueKeyNotInSelect()
```
wiarlawd, line 1156: O Rly? Maybe missingUniqueKey would be OK?

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -1270,6 +1271,22 @@ public void testInitVerifyColumnNames_metadataColumns() throws Exception {
   }
 
   @Test
+  public void testInitVerifyColumnNames_metadataColumnAlias() throws Exception {
```
wiarlawd, line 1274: A negative test would be good, too, that we can't find "other".

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -1916,6 +1962,27 @@ public void testGetDocContent() throws Exception {
   }
 
   @Test
+  public void testGetDocContent_noUniqueKeyFieldsInSelect() throws Exception {
```
wiarlawd, line 1984: Just to really mess up the purpose of this CL, what about a test that uses the wrong column in singleDocContentSql, just to clarify that we can't match column names when substituting parameters. That is, have 2 rows, with id = 1, notId = 0, and id = 2, notId = 1, and use a docId value of 1 but with notId = ? in the query. It doesn't test anything in our code, it just confirms intent and functionality in the way the queries are written.

wiarlawd (commented):
A couple of suggested tests.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: I don't think that is better, and perhaps even worse, since it sounds like it should be an error.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: Done.

wiarlawd (approved):

PR #85, opened 2017-07-27
Changes to pass most tests on SQL Server.

srinicodebytes:


BrettMichaelJohnson (approved):

wiarlawd (commented):
Please add a description that quantifies how many tests pass and fail on SQL Server, and add a note that you mostly added sizes to VARCHAR types (which is required by the SQL standard).

PR #86, opened 2017-07-27
Fix several existing tests that were broken on MySQL

BrettMichaelJohnson:
- MySQL does not support CLOBs or ARRAYs
- Specify nullable timestamps when nulling them
- Specify millisecond resolution for timestamps when required
- MySQL columns are case-sensitive

Tests that remain broken:
 - all that use dateadd()

### DIFF
--- a/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
+++ b/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
@@ -805,9 +805,10 @@ public class DatabaseAdaptorTest {
 
   @Test
   public void testInitMetadataColumns_trueBlank() throws Exception {
+    // Note: MySQL is case-preserving in column names on table creation.
     executeUpdate(
         "create table data(ID int, FOO varchar(20), BAR varchar(20))");
-    executeUpdate("insert into data(ID, FOO, BAR) "
+    executeUpdate("insert into data(id, foo, bar) "
                   + "values(1, 'fooVal', 'barVal')");
 
     Map<String, String> moreEntries = new HashMap<String, String>();
@@ -815,10 +816,10 @@ public class DatabaseAdaptorTest {
     moreEntries.put("db.metadataColumns", "");
     // Required for validation, but not specific to this test.
     moreEntries.put("db.modeOfOperation", "rowToText");
-    moreEntries.put("db.uniqueKey", "ID:int");
+    moreEntries.put("db.uniqueKey", "id:int");
     moreEntries.put("db.everyDocIdSql", "select ID from data");
     moreEntries.put("db.singleDocContentSql",
-        "select * from data where ID = ?");
+        "select * from data where id = ?");
 
     DatabaseAdaptor adaptor = getObjectUnderTest(moreEntries);
     ResultSet resultSet = executeQueryAndNext("select * from data");
--- a/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
+++ b/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
@@ -806,8 +806,8 @@ public class DatabaseAdaptorTest {
   @Test
   public void testInitMetadataColumns_trueBlank() throws Exception {
     executeUpdate(
-        "create table data(id int, foo varchar(20), bar varchar(20))");
-    executeUpdate("insert into data(id, foo, bar) "
+        "create table data(ID int, FOO varchar(20), BAR varchar(20))");
+    executeUpdate("insert into data(ID, FOO, BAR) "
                   + "values(1, 'fooVal', 'barVal')");
 
     Map<String, String> moreEntries = new HashMap<String, String>();
@@ -815,10 +815,10 @@ public class DatabaseAdaptorTest {
     moreEntries.put("db.metadataColumns", "");
     // Required for validation, but not specific to this test.
     moreEntries.put("db.modeOfOperation", "rowToText");
-    moreEntries.put("db.uniqueKey", "id:int");
-    moreEntries.put("db.everyDocIdSql", "select id from data");
+    moreEntries.put("db.uniqueKey", "ID:int");
+    moreEntries.put("db.everyDocIdSql", "select ID from data");
     moreEntries.put("db.singleDocContentSql",
-        "select * from data where id = ?");
+        "select * from data where ID = ?");
 
     DatabaseAdaptor adaptor = getObjectUnderTest(moreEntries);
     ResultSet resultSet = executeQueryAndNext("select * from data");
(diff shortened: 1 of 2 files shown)


BrettMichaelJohnson:
MySQL does have a DATE_ADD function with a different syntax, so we can probably have a function that generates the correct syntax per DB.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -803,19 +806,19 @@ public void testInitMetadataColumns_falseBlank() throws Exception {
   @Test
   public void testInitMetadataColumns_trueBlank() throws Exception {
     executeUpdate(
-        "create table data(id int, foo varchar(20), bar varchar(20))");
-    executeUpdate("insert into data(id, foo, bar) "
+        "create table data(ID int, FOO varchar(20), BAR varchar(20))");
```
wiarlawd: Why are you making this change?

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2221,7 +2224,7 @@ public void testMetadataColumns_timeNull() throws Exception {
 
   @Test
   public void testMetadataColumns_timestamp() throws Exception {
-    executeUpdate("create table data(id integer, col timestamp)");
+    executeUpdate("create table data(id integer, col timestamp(3))");
```
wiarlawd, line 2229: I'm not sure what to do here. The TIMESTAMP type in MySQL is wonky (it stores milliseconds since the epoch internally). DATETIME seems more standard. I think we should plan on doing swizzling in JdbcFixture, maybe with a specialized createTable method instead of the generic executeUpdate, and swizzle from standard SQL as much as possible.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2246,7 +2249,7 @@ public void testMetadataColumns_timestamp() throws Exception {
 
   @Test
   public void testMetadataColumns_timestampNull() throws Exception {
-    executeUpdate("create table data(id integer, col timestamp)");
+    executeUpdate("create table data(id integer, col timestamp null)");
```
wiarlawd: s/null/default null/ may be more portable.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2269,6 +2272,7 @@ public void testMetadataColumns_timestampNull() throws Exception {
 
   @Test
   public void testMetadataColumns_clob() throws Exception {
+    assumeFalse("MySQL does not support clobs", is(MYSQL));
```
wiarlawd, line 2277: This seems like it's going to get unwieldy fast. Have you thought about running on just H2 and SQL Server, or not on one or more DBs? "assumeTrue(not(MYSQL))" might be more readable, as might "assumeThat(DATABASE, not(MYSQL))", and the latter is much more general as well. I wonder if the textual comments will be more noisy than useful.

wiarlawd (commented):
Only blocking comment is the question about uppercasing identifiers.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: MySQL is case preserving in column names.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: timestamp stores timestamps rounded to seconds.
timestamp(3) stores milliseconds.
timestamp(6) stores microseconds.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: The messages show up in the test output as:
Testcase: testMetadataColumns_array took 0.003 sec
    	SKIPPED: MySQL does not support arrays

I find it useful.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -803,19 +806,19 @@ public void testInitMetadataColumns_falseBlank() throws Exception {
...
                   + "values(1, 'fooVal', 'barVal')");
 
     Map<String, String> moreEntries = new HashMap<String, String>();
     moreEntries.put("db.includeAllColumnsAsMetadata", "true");
     moreEntries.put("db.metadataColumns", "");
     // Required for validation, but not specific to this test.
     moreEntries.put("db.modeOfOperation", "rowToText");
-    moreEntries.put("db.uniqueKey", "id:int");
-    moreEntries.put("db.everyDocIdSql", "select id from data");
+    moreEntries.put("db.uniqueKey", "ID:int");
```
wiarlawd: This really should not change.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
wiarlawd: That combined with the case-sensitivity of keys in Metadata.equals, which is the real issue. Please add a comment about that. Also, if it's just case-preserving, why do you need uppercase everywhere, and which one actually matters (the table definition or the query)?

BrettMichaelJohnson: It is the table definition that matters. I will revert the rest of the changes.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: reverted

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -805,20 +805,21 @@ public void testInitMetadataColumns_falseBlank() throws Exception {
 
   @Test
   public void testInitMetadataColumns_trueBlank() throws Exception {
+    // Note: MySQL is case-preserving in column names on table creation.
```
wiarlawd: s/\./ and Metadata.equals is case-sensitive./

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -813,7 +817,7 @@ public void testInitMetadataColumns_trueBlank() throws Exception {
     // Required for validation, but not specific to this test.
     moreEntries.put("db.modeOfOperation", "rowToText");
     moreEntries.put("db.uniqueKey", "id:int");
-    moreEntries.put("db.everyDocIdSql", "select id from data");
+    moreEntries.put("db.everyDocIdSql", "select ID from data");
```
wiarlawd: Not reverted.

wiarlawd (approved):
LGTM with comments.

PR #88, opened 2017-08-02
Fix more tests for MySQL

BrettMichaelJohnson:
- Many instances of varchar - add size
- Exclude tests of unsupported types (clob, array)
- Uppercase column names in tests that needed to preserve their case
- Added tests for LONGVARBINARY
- Use VARBINARY, rather than BINARY in some tests

Unrelated changes:
- UniqueKeyTest.testUniqueElementsRoundTrip() was executing the wrong query
- I could not run all the tests as I was running out of DB connections.
  I updated JdbcFixture.executeQuery() to add its connection to the openObjects
  deque.
- JdbcFixture.dropAllObjects() now empties the openObjects deque.

To be done in a second CL:
- A bunch of tests still leak connections, so I will sweep through all those
  that do try (PreparedStatement ps = getConnection().prepareStatement()) {
  and add the connection to the try-with-resources.

### DIFF
--- a/build.properties.sample
+++ b/build.properties.sample
@@ -20,7 +20,8 @@ build.bootclasspath = ${jdk7.home}/jre/lib/rt.jar\
 # Supported databases: H2, MySQL, Oracle, SQLServer
 # test.jdbc.database = MySQL
 # test.jdbc.driver = com.mysql.jdbc.Driver
-# test.jdbc.url = jdbc:mysql://address=(useSSL=false)(protocol=tcp)(host=localhost)(port=3306)/test?sessionVariables=sql_mode='ANSI_QUOTES'
+# test.jdbc.url = jdbc:mysql://address=(useSSL=false)(protocol=tcp)(host=localhost)(port=3306)/test\
+  ?sessionVariables=sql_mode='ANSI_QUOTES'&useLegacyDatetimeCode=false&serverTimezone=US/Pacific
 # test.jdbc.user =
 # test.jdbc.password =
 #
--- a/build.properties.sample
+++ b/build.properties.sample
@@ -21,7 +21,7 @@ build.bootclasspath = ${jdk7.home}/jre/lib/rt.jar\
 # test.jdbc.database = MySQL
 # test.jdbc.driver = com.mysql.jdbc.Driver
 # test.jdbc.url = jdbc:mysql://address=(useSSL=false)(protocol=tcp)(host=localhost)(port=3306)/test\
-  ?sessionVariables=sql_mode='ANSI_QUOTES'&useLegacyDatetimeCode=false&serverTimezone=US/Pacific
+#     ?sessionVariables=sql_mode='ANSI_QUOTES'&useLegacyDatetimeCode=false&serverTimezone=US/Pacific
 # test.jdbc.user =
 # test.jdbc.password =
 #
--- a/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
+++ b/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
@@ -2283,8 +2283,6 @@ public class DatabaseAdaptorTest {
 
   @Test
   public void testMetadataColumns_clob() throws Exception {
-    // CLOB is TEXT in MySQL
-    // NCLOB shows up as CLOB in H2
     String content = "Hello World";
     executeUpdate("create table data(id int, content clob)");
     String sql = "insert into data(id, content) values (1, ?)";
--- a/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
+++ b/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
@@ -2260,7 +2260,7 @@ public class DatabaseAdaptorTest {
 
   @Test
   public void testMetadataColumns_timestampNull() throws Exception {
-    executeUpdate("create table data(id integer, col timestamp default null)");
+    executeUpdate("create table data(id integer, col timestamp)");
     executeUpdate("insert into data(id) values(1001)");
 
     Map<String, String> moreEntries = new HashMap<String, String>();
--- a/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
+++ b/test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
@@ -16,7 +16,7 @@ package com.google.enterprise.adaptor.database;
 
 import static com.google.enterprise.adaptor.DocIdPusher.Record;
 import static com.google.enterprise.adaptor.Principal.DEFAULT_NAMESPACE;
-import static com.google.enterprise.adaptor.database.JdbcFixture.Database.MYSQL;
+import static com.google.enterprise.adaptor.database.JdbcFixture.Database.H2;
 import static com.google.enterprise.adaptor.database.JdbcFixture.executeQuery;
 import static com.google.enterprise.adaptor.database.JdbcFixture.executeQueryAndNext;
 import static com.google.enterprise.adaptor.database.JdbcFixture.executeUpdate;
@@ -36,7 +36,7 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
-import static org.junit.Assume.assumeFalse;
+import static org.junit.Assume.assumeTrue;
 
 import com.google.enterprise.adaptor.Acl;
 import com.google.enterprise.adaptor.AuthnIdentity;
@@ -2283,7 +2283,7 @@ public class DatabaseAdaptorTest {
 
   @Test
   public void testMetadataColumns_clob() throws Exception {
-    assumeFalse("MySQL does not support clobs", is(MYSQL));
+    // CLOB is TEXT in MySQL
     // NCLOB shows up as CLOB in H2
     String content = "Hello World";
     executeUpdate("create table data(id int, content clob)");
@@ -2314,7 +2314,6 @@ public class DatabaseAdaptorTest {
 
   @Test
   public void testMetadataColumns_clobNull() throws Exception {
-    assumeFalse("MySQL does not support clobs", is(MYSQL));
     executeUpdate("create table data(id int, content clob)");
     executeUpdate("insert into data(id) values (1)");
 
@@ -2397,7 +2396,7 @@ public class DatabaseAdaptorTest {
 
   @Test
   public void testMetadataColumns_array() throws Exception {
-    assumeFalse("MySQL does not support arrays", is(MYSQL));
+    assumeTrue("ARRAY type not supported", is(H2));
     String[] content = { "hello", "world" };
     executeUpdate("create table data(id int, content array)");
     String sql = "insert into data(id, content) values (1, ?)";
@@ -2430,7 +2429,7 @@ public class DatabaseAdaptorTest {
 
   @Test
   public void testMetadataColumns_arrayNull() throws Exception {
-    assumeFalse("MySQL does not support arrays", is(MYSQL));
+    assumeTrue("ARRAY type not supported", is(H2));
     executeUpdate("create table data(id int, content array)");
     executeUpdate("insert into data(id) values (1)");
 
--- a/test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
+++ b/test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
@@ -446,7 +446,7 @@ public class ResponseGeneratorTest {
     executeUpdate("create table data(id int, content timestamp)");
     String sql = "insert into data(id, content) values (1, ?)";
     try (PreparedStatement ps = getConnection().prepareStatement(sql)) {
-        ps.setTimestamp(1, new Timestamp(0L));
+      ps.setTimestamp(1, new Timestamp(0L));
       assertEquals(1, ps.executeUpdate());
     }
 
--- a/test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
+++ b/test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
@@ -14,6 +14,7 @@
 
 package com.google.enterprise.adaptor.database;
 
+import static com.google.enterprise.adaptor.database.JdbcFixture.Database.H2;
 import static com.google.enterprise.adaptor.database.JdbcFixture.Database.MYSQL;
 import static com.google.enterprise.adaptor.database.JdbcFixture.executeQueryAndNext;
 import static com.google.enterprise.adaptor.database.JdbcFixture.executeUpdate;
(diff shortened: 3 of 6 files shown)


build.properties.sample
```diff
@@ -20,7 +20,8 @@ build.bootclasspath = ${jdk7.home}/jre/lib/rt.jar\
 # Supported databases: H2, MySQL, Oracle, SQLServer
 # test.jdbc.database = MySQL
 # test.jdbc.driver = com.mysql.jdbc.Driver
-# test.jdbc.url = jdbc:mysql://address=(useSSL=false)(protocol=tcp)(host=localhost)(port=3306)/test?sessionVariables=sql_mode='ANSI_QUOTES'
+# test.jdbc.url = jdbc:mysql://address=(useSSL=false)(protocol=tcp)(host=localhost)(port=3306)/test\
+  ?sessionVariables=sql_mode='ANSI_QUOTES'&useLegacyDatetimeCode=false&serverTimezone=US/Pacific
```
wiarlawd: Do we need to hard-code the time zone? Does this actually work if you're in a different time zone? Where was the PDT coming from, Java or MySQL?

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -156,6 +161,7 @@ public Object invoke(Object proxy, Method method, Object[] args)
 
   @Test
   public void testRowToText_multipleTypes() throws Exception {
+    assumeFalse("MySQL does not support arrays or clobs", is(MYSQL));
```
wiarlawd: Instead of splintering this completely by database type, what about keeping this as a full test perhaps just on H2 (SQL Server doesn't have arrays, either), and change your new test to run everywhere with the common subset of basic types as you have it?

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -190,9 +196,41 @@ public void testRowToText_multipleTypes() throws Exception {
...
+        + "1,true,\"hello, world\",2007-08-09,12:34:56,2007-08-09 12:34:56.7\n";
+    assertEquals(golden, bar.baos.toString(UTF_8.name()));
+    assertEquals(messages.toString(), 1, messages.size());
+  }
+
+  @Test
   public void testRowToText_null() throws Exception {
     executeUpdate(
-        "create table data (id integer, this varchar(20), that clob)");
+        "create table data (id integer, this varchar(20), that timestamp)");
```
wiarlawd: I would prefer a stream type in place of CLOB, rather than another scalar.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -202,7 +240,10 @@ public void testRowToText_null() throws Exception {
 
     ResultSet rs = executeQueryAndNext("select * from data");
     resgen.generateResponse(rs, response);
-    String golden = "DATA,DATA,DATA\nID,THIS,THAT\n1,,\n";
+    String golden =
```
wiarlawd: Why not just switch to equalsIgnoreCase, or uppercase the result with assertEquals?

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -290,7 +333,7 @@ public void testContentColumn_incorrectColumnName() throws Exception {
   @Test
   public void testContentColumn_varbinary() throws Exception {
     String content = "hello world";
-    executeUpdate("create table data(id int, content varbinary)");
+    executeUpdate("create table data(id int, content varbinary(20))");
```
wiarlawd, line 324: So Srinivas missed these due, presumably, to the Ant haltonfailure="yes"?

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -415,7 +460,7 @@ public void testContentColumn_invalidType() throws Exception {
     executeUpdate("create table data(id int, content timestamp)");
     String sql = "insert into data(id, content) values (1, ?)";
     try (PreparedStatement ps = getConnection().prepareStatement(sql)) {
-      ps.setTimestamp(1, new Timestamp(0L));
+      ps.setTimestamp(1, new Timestamp(new Date().getTime()));
```
wiarlawd: Why? I don't want to just litter the tests with random things that we discover. We should document them, in JdbcFixture perhaps for global things (always put a size on VARCHAR, use uppercase names in create table statements, etc.), and locally for test-specific things.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -147,13 +151,15 @@ public void testInteger_null() throws Exception {
...
-    executeUpdate("create table data(colname boolean)");
+    executeUpdate("create table data(COLNAME boolean)");
     executeUpdate("insert into data(colname) values(true)");
     final String golden = ""
         + "<database>"
         + "<table>"
         + "<table_rec>"
-        + "<COLNAME SQLType=\"BOOLEAN\">true</COLNAME>"
+        + "<COLNAME SQLType=\""
+        + (is(MYSQL) ? "BIT" : "BOOLEAN")
```
wiarlawd: Maybe add a JdbcFixture.BOOLEAN?

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -417,9 +425,11 @@ public void testTime_null() throws Exception {
 
   @Test
   public void testTimestamp() throws Exception {
-    executeUpdate("create table data(colname timestamp)");
+    String timestamp =
+        is(MYSQL) ? "2004-10-06 09:15:30" : "2004-10-06T09:15:30";
```
wiarlawd: Maybe we could use a real database instead. This is obscene.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -664,7 +680,72 @@ public void testBlob_null() throws Exception {
   }
 
   @Test
+  public void testLongvarbinary() throws Exception {
```
wiarlawd: These tests are in the same order as the switch in TupleReader.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -730,6 +813,7 @@ public void testOther() throws Exception {
 
   @Test
   public void testOther_invalidXmlChars() throws Exception {
+    assumeFalse("MySQL does not support type other", is(MYSQL));
```
wiarlawd: This use of OTHER is specific to H2.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -787,7 +872,7 @@ public void testOther_null() throws Exception {
 
   @Test
   public void testMultipleTypes() throws Exception {
-    executeUpdate("create table data(id integer, name varchar, modified date)");
+    executeUpdate("create table data(ID integer, NAME varchar(20), MODIFIED date)");
```
wiarlawd: 80 cols

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -317,15 +319,15 @@ public void testPreparingRetrieval_wrongColumnCount() throws SQLException {
   @Test
   public void testPreparingRetrieval() throws SQLException {
     executeUpdate("create table data("
-        + "numnum int, strstr varchar, timestamp timestamp, date date, "
-        + "time time, long bigint)");
+        + "numnum int, strstr varchar(20), tymestamp timestamp(3), date date, "
```
wiarlawd, line 322: Let's try to come up with a consistent scheme, rather than be random. This is another place we were lazy and non-standard. SQL keywords are reserved. I don't like adding "col" to each (or many) names. Curious that "timestamp" is an error but "time" is not. I'm not entirely opposed to making the first vowel "y", silly though it is, or adding a prefix or suffix character (0 or z or c?). I would stick to alphanumerics.

wiarlawd (changes requested):

build.properties.sample
BrettMichaelJohnson: These are example URLs, and as you saw, I needed to disambiguate the timezone. Specifying the timezone is probably not required if the user is in an unambiguous timezone.  I don't know if JDBC is pulling the default timezone from the DB server or from system properties.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
BrettMichaelJohnson: I broke out array into a different test.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
BrettMichaelJohnson: Restored CLOB

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
BrettMichaelJohnson: probably.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
BrettMichaelJohnson: This can go back.  The problem with 0L on MySQL was the timezone shift put us 8 hours before the Unix epoch.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
BrettMichaelJohnson: Fixed in JdbcFixture

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
BrettMichaelJohnson: Moved to below testBinary*

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
BrettMichaelJohnson: Made it H2 specific.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
BrettMichaelJohnson: Timestamp isn't an issue with the SQL, but it interfered with my create table rewriting of timestamp in JdbcFixture. However the "long" on the next line did run afoul of MySQL, which considers it a modifier of other data types.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -507,7 +568,7 @@ public void testTimestamp_threadSafe() throws Throwable {
           @Override public void run() {
             try {
               ResultSet rs = executeQueryAndNext(
-                  "select colname from data where thread = " + tt);
+                  "select COLNAME from data where thread = " + tt);
```
wiarlawd, line 571: I thought this wasn't needed.

build.properties.sample
wiarlawd: It looks like this line isn't commented out.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2283,7 +2283,7 @@ public void testMetadataColumns_timestampNull() throws Exception {
 
   @Test
   public void testMetadataColumns_clob() throws Exception {
-    assumeFalse("MySQL does not support clobs", is(MYSQL));
+    // CLOB is TEXT in MySQL
```
wiarlawd: I don't really get the point of these comments here. It's not specific to this test, and is part of the swizzling in JdbcFixture. I also wasn't sure about the NCLOB reference, since there is no NCLOB anywhere in the tests.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -155,19 +160,43 @@ public Object invoke(Object proxy, Method method, Object[] args)
...
+  }
+
+  @Test
   public void testRowToText_multipleTypes() throws Exception {
     executeUpdate("create table data ("
         + "intcol integer, booleancol boolean, charcol varchar(20),"
-        + " datecol date, timecol time, timestampcol timestamp,"
-        + " clobcol clob, blobcol blob, arraycol array)");
+        + " datecol date, timecol time, tymestampcol timestamp(3),"
+        + " charlob clob, blobcol blob)");
```
wiarlawd, line 192: Why rename clobcol here?

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -415,7 +446,7 @@ public void testContentColumn_invalidType() throws Exception {
     executeUpdate("create table data(id int, content timestamp)");
     String sql = "insert into data(id, content) values (1, ?)";
     try (PreparedStatement ps = getConnection().prepareStatement(sql)) {
-      ps.setTimestamp(1, new Timestamp(0L));
+        ps.setTimestamp(1, new Timestamp(0L));
```
wiarlawd: Undo indentation.

wiarlawd (approved):
LGTM with nits.

build.properties.sample
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
BrettMichaelJohnson: Nuked them.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
BrettMichaelJohnson: It interfered with the clob swizzling in JdbcFixture.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
BrettMichaelJohnson: Done.

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
BrettMichaelJohnson: It was.

PR #92, opened 2017-08-08
Run all tests even when some fail.

wiarlawd:
Use haltonfailure="no" to run all test classes, but we still need to
fail in the test target in order to stop the build for the dist
target.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -257,8 +257,10 @@ lib/plexi submodule or add the the command line argument
     description="Run instrumented tests and generate coverage report"/>
 
   <target name="test" depends="build" description="Run JUnit tests">
-    <junit printsummary="yes" haltonfailure="no" failureproperty="test.failure"
-           forkmode="once" fork="true" dir="${basedir}" maxmemory="512m">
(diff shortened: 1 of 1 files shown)


srinicodebytes (approved):

BrettMichaelJohnson (approved):

build.xml
```diff
@@ -257,8 +257,10 @@ lib/plexi submodule or add the the command line argument
     description="Run instrumented tests and generate coverage report"/>
 
   <target name="test" depends="build" description="Run JUnit tests">
-    <junit printsummary="yes" haltonfailure="no" failureproperty="test.failure"
-           forkmode="once" fork="true" dir="${basedir}" maxmemory="512m">
+    <property name="build.haltonfailure" value="off"/>
```
BrettMichaelJohnson: In the DB adaptor, we started using properties named test.* for the tests.  How about test.haltonfailure?

wiarlawd: I renamed it on purpose. The test.* properties are sent through to the tests, and these properties are purely for the build.

PR #93, opened 2017-08-17
Pass all tests against SQL Server.

wiarlawd:
* SQL NULL cannot be read as a datetime2 without a cast.
* SQL CHAR is padded with trailing spaces to the given size.
* Use bit for BOOLEAN.
* Use varbinary(max) and varchar(max) for BLOB and CLOB.
* Avoid ts escapes, which produce a datetime rather than a datetime2.
* Reintroduce the table names in rowtotext mode, and use empty
  strings for SQL Server.
* Use empty byte array rather than empty string for binary data.

Other changes:

* Revamp the executeUpdate swizzling to use \b and replaceAll.

### DIFF
--- a/test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
+++ b/test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
@@ -24,6 +24,7 @@ import static com.google.enterprise.adaptor.database.Logging.captureLogMessages;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Locale.US;
 import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.endsWith;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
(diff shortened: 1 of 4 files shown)


test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -211,8 +210,10 @@ public void testRowToText_multipleTypes() throws Exception {
     captureLogMessages(ResponseGenerator.class,
         "Column type not supported for text", messages);
     resgen.generateResponse(rs, response);
-    String golden = "data,data,data,data,data,data,data\n"
-        + "intcol,booleancol,charcol,datecol,timecol,tymestampcol,charlob\n"
+    String tables = is(SQLSERVER)
+        ? ",,,,,," : "data,data,data,data,data,data,data";
+    String golden = tables + "\n"
+        + "intcol,booleancol,charcol,datecol,timecol,tymestampcol,clobcol\n"
```
BrettMichaelJohnson: Since you put back clobcol, you might as well put back timestampcol.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
```diff
@@ -253,11 +255,10 @@ public void testRowToText_specialCharacters() throws Exception {
 
     ResultSet rs = executeQueryAndNext("select * from data");
     resgen.generateResponse(rs, response);
-    String golden =
-        (is(MYSQL) ? "id,name,quote\n"
-                   : "ID,NAME,QUOTE\n")
+    String table = is(SQLSERVER) ? ",," : "DATA,DATA,DATA";
```
BrettMichaelJohnson: This is still going to break on MySQL, where the table name comes back in lower case, depending on the filesystem used.

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
wiarlawd: Done

test/com/google/enterprise/adaptor/database/ResponseGeneratorTest.java
wiarlawd: Skipped the table name comparison in this test, since it appears elsewhere. Just simplified the test for MySQL (and SQL Server).

BrettMichaelJohnson (approved):

PR #95, opened 2017-08-19
Pass all but one test against Oracle.

wiarlawd:
The exception is DatabaseAdaptorTest.testInitUniqueKeyOptionalTypes,
which uncovered b/64541388 and will be fixed separately.

* The DATE data type has both date and time components. Add t and d
  methods to JdbcFixture to mirror the resulting values from the
  JDBC escapes of the same names.
* Empty character and binary values as treated as null.
* The INTEGER data type is an alias for NUMBER(38) and returns as
  Types.NUMERIC rather than Types.INTEGER.
* BIT and BOOLEAN are not supported.
* Use integer for BIGINT.
* Use date for TIME (in some contexts).
* Use raw and long raw for VARBINARY and LONGVARBINARY.

### DIFF
--- a/test/com/google/enterprise/adaptor/database/TupleReaderTest.java
+++ b/test/com/google/enterprise/adaptor/database/TupleReaderTest.java
@@ -126,7 +126,7 @@ public class TupleReaderTest {
         + "<table>"
         + "<table_rec>"
         + "<COLNAME SQLType=\""
-        + (is(ORACLE) ? "NUMERIC" : "INTEGER")
+        + JdbcFixture.INTEGER
         + "\">17</COLNAME>"
         + "</table_rec>"
         + "</table>"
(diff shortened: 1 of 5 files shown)


test/com/google/enterprise/adaptor/database/TupleReaderTest.java
```diff
@@ -124,7 +125,9 @@ public void testInteger() throws Exception {
         + "<database>"
         + "<table>"
         + "<table_rec>"
-        + "<COLNAME SQLType=\"INTEGER\">17</COLNAME>"
+        + "<COLNAME SQLType=\""
+        + (is(ORACLE) ? "NUMERIC" : "INTEGER")
```
BrettMichaelJohnson: Should you handle this the same way we did JdbcFixture.BOOLEAN?

BrettMichaelJohnson (approved):
LGTM++

BrettMichaelJohnson (approved):

test/com/google/enterprise/adaptor/database/TupleReaderTest.java
wiarlawd: Done

PR #96, opened 2017-08-19
Support unique key columns of type NUMERIC (b/64541388).

wiarlawd:
Tested against H2, Oracle, and SQL Server.

### DIFF
--- a/test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
+++ b/test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
@@ -14,9 +14,11 @@
 
 package com.google.enterprise.adaptor.database;
 
+import static com.google.enterprise.adaptor.database.JdbcFixture.Database.H2;
 import static com.google.enterprise.adaptor.database.JdbcFixture.Database.ORACLE;
 import static com.google.enterprise.adaptor.database.JdbcFixture.executeQueryAndNext;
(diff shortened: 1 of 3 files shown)


test/com/google/enterprise/adaptor/database/UniqueKeyTest.java
```diff
@@ -248,16 +312,18 @@ public void testProcessingDocId() throws Exception {
 
   @Test
   public void testProcessingDocId_allTypes() throws Exception {
-    executeUpdate("create table data(c1 integer, c2 bigint, c3 varchar(20), "
-        + "c4 date, c5 time, c6 timestamp)");
-    executeUpdate("insert into data(c1, c2, c3, c4, c5, c6) "
-        + "values (123, 4567890, 'foo', "
+    executeUpdate("create table data(c1 integer, c2 bigint, c3 numeric(9,2), "
```
BrettMichaelJohnson, line 315: Will these work on H2?

wiarlawd: Yes. It's just the auto-detect that's broken on H2, because it returns Types.DECIMAL, which we don't want to support (it's inexact). Not sure why they chose that for BigDecimal, maybe just got confused by the names. Numeric types work just fine as long as you declare them as such.

BrettMichaelJohnson (approved):

PR #107, opened 2017-09-29
Trigger more javac lint warnings and fix them.

wiarlawd:
This undoes the syntactic AutoCloseable changes in commit 6027217b,
but not the semantic changes made there, flushing unsent records.


PapaJoltOmega (approved):
Thank you

PR #110, opened 2017-11-28
DB Fix bug#67471678. Include ACLs in the db.singleDocContentSql

srinicodebytes:


src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -274,9 +274,9 @@ public void init(AdaptorContext context) throws Exception {
     if (!isNullOrEmptyString(cfg.getValue("db.aclSql"))) {
       aclSql = cfg.getValue("db.aclSql");
       log.config("acl sql: " + aclSql); 
-      aclPrincipalDelimiter = cfg.getValue("db.aclPrincipalDelimiter");
-      log.config("aclPrincipalDelimiter: '" + aclPrincipalDelimiter + "'");
     }
+    aclPrincipalDelimiter = cfg.getValue("db.aclPrincipalDelimiter");
```
PapaJoltOmega: add comment
// we might have delimiter and no aclSql as ACLs might come from singleDocContentSql rather than aclSql

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -697,18 +700,27 @@ private Acl getAcl(Connection conn, String uniqueId) throws SQLException {
...
+        return Acl.EMPTY;
+      } else {
+        return buildAcl(rs, aclPrincipalDelimiter, aclNamespace,
+            ResultSetNext.CONTINUE_NEXT);
+      }
     }
   }
   
   @VisibleForTesting
   static Acl buildAcl(ResultSet rs, String delim, String namespace)
```
wiarlawd: Who calls this overload?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -678,6 +678,9 @@ public void getDocContent(Request req, Response resp) throws IOException {
       // Generate Acl if aclSql is provided.
       if (aclSql != null) {
         resp.setAcl(getAcl(conn, id.getUniqueId()));
+      } else {
+        resp.setAcl(buildAcl(rs, aclPrincipalDelimiter, aclNamespace,
```
PapaJoltOmega: add comment 
// see if ACL came as part of result to singleDocContentSql query

wiarlawd: This seems like it's going to set the empty ACL (an explicitly constructed one, rather than Acl.EMPTY) on a document that should be public.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -810,14 +825,22 @@ private Connection makeNewConnection() throws SQLException {
   private PreparedStatement getDocFromDb(Connection conn,
       String uniqueId) throws SQLException {
     PreparedStatement st = conn.prepareStatement(singleDocContentSql);
-    uniqueKey.setContentSqlValues(st, uniqueId);  
+    uniqueKey.setContentSqlValues(st, uniqueId);
```
wiarlawd: Restore trailing whitespace on this unmodified line.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -740,6 +752,9 @@ static Acl buildAcl(ResultSet rs, String delim, String namespace)
         denyGroups.addAll(getGroupPrincipalsFromResultSet(rs,
             GsaSpecialColumns.GSA_DENY_GROUPS, delim, namespace));
       }
+      if (rsNext == ResultSetNext.STOP_NEXT) {
```
PapaJoltOmega: Please help me understand the constraint that ACL coming from singleDocContentSql imposes that requires this statement.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
PapaJoltOmega: has something changed on this line?  github shows it as changed.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -810,14 +825,22 @@ private Connection makeNewConnection() throws SQLException {
...
 
   private PreparedStatement getAclFromDb(Connection conn,
       String uniqueId) throws SQLException {
-    PreparedStatement st = conn.prepareStatement(aclSql);
+    PreparedStatement st;
+    if (aclSql == null) {
+       st = conn.prepareStatement(singleDocContentSql);
+    } else {
+       st = conn.prepareStatement(aclSql);
+    }
```
PapaJoltOmega: indent looks 1off

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -1085,6 +1108,11 @@ private PreparedStatement getUpdateStreamFromDb(Connection conn)
     GSA_TIMESTAMP;
   }
 
+  enum ResultSetNext {
+    CONTINUE_NEXT,
+    STOP_NEXT,
+  }
+
```
PapaJoltOmega: How about
PROCESS_ALL_RESULTS
PROCESS_ONE_RESULT

Am I understanding the semantics?

PapaJoltOmega (commented):
Thank you.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -810,14 +825,22 @@ private Connection makeNewConnection() throws SQLException {
...
     log.log(Level.FINER, "about to get doc: {0}",  uniqueId);
     return st;
   }
 
   private PreparedStatement getAclFromDb(Connection conn,
       String uniqueId) throws SQLException {
-    PreparedStatement st = conn.prepareStatement(aclSql);
+    PreparedStatement st;
+    if (aclSql == null) {
+       st = conn.prepareStatement(singleDocContentSql);
```
wiarlawd: It seems worth a comment here that this only happens when called from isUserAuthorized. Also, we will execute db.singleDocContentSql twice in that case. I suppose that's inevitable, but did you think about trying to avoid that?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -1085,6 +1108,11 @@ private PreparedStatement getUpdateStreamFromDb(Connection conn)
     GSA_TIMESTAMP;
   }
 
+  enum ResultSetNext {
+    CONTINUE_NEXT,
```
wiarlawd: What do you think about ALL_ROWS, ONE_ROW?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Thx, fixed it.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -417,6 +418,44 @@ public void testAclSqlResultSetHasOneRecord() throws SQLException {
...
 
+  @Test
+  public void testSingleDocSqlResultSetHasOneRecord() throws SQLException {
+    executeUpdate("create table data("
+        + "id int, col1 varchar(20),"
+        + GsaSpecialColumns.GSA_PERMIT_GROUPS + " varchar(20),"
+        + GsaSpecialColumns.GSA_PERMIT_USERS + " varchar(20),"
+        + GsaSpecialColumns.GSA_DENY_GROUPS + " varchar(20),"
+        + GsaSpecialColumns.GSA_DENY_USERS + " varchar(20))");
+    executeUpdate("insert into data("
```
wiarlawd, line 438: This test should minimally have 2 rows as a test of STOP_NEXT.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -417,6 +418,44 @@ public void testAclSqlResultSetHasOneRecord() throws SQLException {
...
+        .setPermitGroups(Arrays.asList(
+            new GroupPrincipal("pgroup1"),
+            new GroupPrincipal("pgroup2")))
+        .setDenyGroups(Arrays.asList(
+            new GroupPrincipal("dgroup2"),
+            new GroupPrincipal("dgroup1")))
+        .build();
+    ResultSet rs = executeQuery("select * from data");
+    rs.next();
+    Acl acl = DatabaseAdaptor.buildAcl(rs, ",", DEFAULT_NAMESPACE,
```
wiarlawd: There are only six tests that use buildAcl instead of getDocContent. I think we should convert all of them, perhaps in a separate CL. Perhaps minimally we could use getDocContent for this new test. We have more logic around db.aclSql now, and I don't like bypassing that in the tests.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: We shouldn't be calling rs.next() when processing ACLs from singleDocContentSql, or else we would be moving to next or end of result set and wouldn't be able to get content from result set.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Looks like I removed the whitespace at the end. Restored it.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Thx, modified these.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Combining your and PJ's suggestion, modified to PROCESS_ALL_ROWS , PROCESS_ONE_ROW.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Thx, fixed it by returning Acl.EMPTY when no acl info in the result

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Thx, added another row

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Agree. Modified this test to call getDocContent instead of buildAcl.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Only tests currently. Moved this to tests.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -723,6 +731,12 @@ static Acl buildAcl(ResultSet rs, String delim, String namespace)
         hasColumn(metadata, GsaSpecialColumns.GSA_PERMIT_GROUPS);
     boolean hasDenyGroups =
         hasColumn(metadata, GsaSpecialColumns.GSA_DENY_GROUPS);
+
+    if (!hasPermitUsers && !hasDenyUsers
+        && !hasPermitGroups && !hasDenyGroups) {
+      return Acl.EMPTY;
```
wiarlawd: Whether implicit or explicit, the empty ACL is wrong in the case of getting ACLs from db.singleDocContentSql. I think we could just add another argument for the default ACL to use here, either null or Acl.EMPTY. These values are paired with the ResultSetNext enum, so we could conceivably collapse them into a single argument, but that feels a little obfuscated.

Since this is a bug, albeit newly introduced, you should add a test for it.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -364,14 +364,21 @@ public void testHasColumn() throws Exception {
         GsaSpecialColumns.GSA_TIMESTAMP));
   }
 
+  private Acl buildAcl(ResultSet rs, String delim, String namespace)
```
wiarlawd, line 369: I would have left this alone with a TODO, since we're planning to replace it entirely, or you could have used executeQueryAndNext instead of executeQuery. Just mentioning other ideas for posterity, since we're planning to sweep this away anyway.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -449,10 +465,20 @@ public void testSingleDocSqlResultSetHasOneRecord() throws SQLException {
...
             new GroupPrincipal("dgroup1")))
         .build();
-    ResultSet rs = executeQuery("select * from data");
-    rs.next();
-    Acl acl = DatabaseAdaptor.buildAcl(rs, ",", DEFAULT_NAMESPACE,
-        ResultSetNext.STOP_NEXT);
+
+    Map<String, String> configEntries = new HashMap<String, String>();
+    configEntries.put("db.uniqueKey", "id:int");
+    configEntries.put("db.everyDocIdSql", "select id from data");
```
wiarlawd: Follow the boilerplate and put db.everyDocIdSql after the "Required for validation, ..." comment.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -449,10 +465,20 @@ public void testSingleDocSqlResultSetHasOneRecord() throws SQLException {
...
+    configEntries.put("db.singleDocContentSql",
+        "select * from data where id = ?");
+    configEntries.put("db.modeOfOperation", "rowToText");
+
+    DatabaseAdaptor adaptor = getObjectUnderTest(configEntries);
+    MockRequest request = new MockRequest(new DocId("1"));
+    RecordingResponse response = new RecordingResponse();
+    adaptor.getDocContent(request, response);
+
+    Acl acl = response.getAcl();
```
wiarlawd: It's more consistent now, but if we're planning to use this as an exemplar, I think it's nicer to just inline this value in the assert.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -811,13 +828,21 @@ private PreparedStatement getDocFromDb(Connection conn,
...
     log.log(Level.FINER, "about to get doc: {0}",  uniqueId);
     return st;
   }
 
   private PreparedStatement getAclFromDb(Connection conn,
       String uniqueId) throws SQLException {
-    PreparedStatement st = conn.prepareStatement(aclSql);
+    PreparedStatement st;
+    if (aclSql == null) {
+      st = conn.prepareStatement(singleDocContentSql);
```
wiarlawd: Forgotten comment here?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -811,13 +828,21 @@ private PreparedStatement getDocFromDb(Connection conn,
       String uniqueId) throws SQLException {
     PreparedStatement st = conn.prepareStatement(singleDocContentSql);
     uniqueKey.setContentSqlValues(st, uniqueId);  
+    if (aclSql == null) {
+      uniqueKey.setAclSqlValues(st, uniqueId);
```
wiarlawd: This seems wrong. The db.aclSqlParameters columns are specific to db.aclSql, and may not be the same in db.singleDocContentSql. Did something fail without this?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -811,13 +828,21 @@ private PreparedStatement getDocFromDb(Connection conn,
...
+    }
     log.log(Level.FINER, "about to get doc: {0}",  uniqueId);
     return st;
   }
 
   private PreparedStatement getAclFromDb(Connection conn,
       String uniqueId) throws SQLException {
-    PreparedStatement st = conn.prepareStatement(aclSql);
+    PreparedStatement st;
```
wiarlawd: Maybe simpler:

    String sql = (aclSql == null) ? singleDocContentSql : aclSql;
    PreparedStatement st = conn.prepareStatement(sql);

wiarlawd (changes requested):

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
PapaJoltOmega: It seems a little redundant that we run this check before calling this function on line 681. I don't, at this time, have a better suggestion. Let's keep an eye out to see if there is a better way to keep the functions orthogonality composable. The current way is fine, but I will keep an eye out to see if there is something better.  It is also possible that you are planning to revisit/refactor in follow up. Thank you

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
wiarlawd: Since this is wrong, too, maybe add a test where doing this would fail, too.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Yes, thx. Done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -740,6 +754,9 @@ static Acl buildAcl(ResultSet rs, String delim, String namespace)
         denyGroups.addAll(getGroupPrincipalsFromResultSet(rs,
             GsaSpecialColumns.GSA_DENY_GROUPS, delim, namespace));
       }
+      if (rsNext == ResultSetNext.PROCESS_ONE_ROW) {
```
wiarlawd: Or maybe just:

} while (rsNext == ResultSetNext.PROCESS_ALL_ROWS && rs.next());

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Deleted the newly added lines, as these are not required. And no test required.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Added comment. Done.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Agree. Keeping as is for now.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done. Added test for it.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -478,8 +478,28 @@ public void testSingleDocSqlResultSetHasTwoRecords() throws Exception {
...
+    configEntries.put("db.singleDocContentSql",
+        "select * from data where id = ?");
+    configEntries.put("db.modeOfOperation", "rowToText");
+
+    DatabaseAdaptor adaptor = getObjectUnderTest(configEntries);
+    MockRequest request = new MockRequest(new DocId("1"));
+    RecordingResponse response = new RecordingResponse();
+    adaptor.getDocContent(request, response);
+
+    assertEquals(golden, response.getAcl());
```
wiarlawd: The use of golden here seems silly. assertNull(response.getAcl()) is more clear.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -413,10 +421,87 @@ public void testAclSqlResultSetHasOneRecord() throws SQLException {
...
+            new GroupPrincipal("pgroup1"),
+            new GroupPrincipal("pgroup2")))
+        .setDenyGroups(Arrays.asList(
+            new GroupPrincipal("dgroup2"),
+            new GroupPrincipal("dgroup1")))
+        .build();
+
+    Map<String, String> configEntries = new HashMap<String, String>();
+    configEntries.put("db.uniqueKey", "id:int");
+    configEntries.put("db.everyDocIdSql", "select id from data");
```
wiarlawd: Missed comment here: put this at the end of the configEntries, after a comment:

    // Required for validation, but not specific to this test.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -817,7 +834,9 @@ private PreparedStatement getDocFromDb(Connection conn,
 
   private PreparedStatement getAclFromDb(Connection conn,
       String uniqueId) throws SQLException {
-    PreparedStatement st = conn.prepareStatement(aclSql);
+    // aclSql will be null when called from isUserAuthorized.
```
wiarlawd: s/will/will only/

wiarlawd (commented):
Nits

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -697,18 +703,20 @@ private Acl getAcl(Connection conn, String uniqueId) throws SQLException {
     try (PreparedStatement stmt = getAclFromDb(conn, uniqueId);
         ResultSet rs = stmt.executeQuery()) {
       log.finer("got acl");
-      return buildAcl(rs, aclPrincipalDelimiter, aclNamespace);
+      boolean hasResult = rs.next();
+      if (!hasResult) {
+        // empty Acl ensures adaptor will mark this document as secure
+        return Acl.EMPTY;
```
PapaJoltOmega: Why move this code from buildAcl?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -697,18 +703,20 @@ private Acl getAcl(Connection conn, String uniqueId) throws SQLException {
     try (PreparedStatement stmt = getAclFromDb(conn, uniqueId);
```
PapaJoltOmega: Perhaps rename getAclFromDb to runAclSqlStatement (to disambiguate that it is not running content sql)?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -678,6 +680,10 @@ public void getDocContent(Request req, Response resp) throws IOException {
       // Generate Acl if aclSql is provided.
       if (aclSql != null) {
         resp.setAcl(getAcl(conn, id.getUniqueId()));
```
PapaJoltOmega: Perhaps rename getAcl to getAclFromAclSql (to disambiguate that it is not running the content sql)?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -723,6 +731,12 @@ static Acl buildAcl(ResultSet rs, String delim, String namespace)
         hasColumn(metadata, GsaSpecialColumns.GSA_PERMIT_GROUPS);
     boolean hasDenyGroups =
         hasColumn(metadata, GsaSpecialColumns.GSA_DENY_GROUPS);
+
+    if (!hasPermitUsers && !hasDenyUsers
+        && !hasPermitGroups && !hasDenyGroups) {
+      return defaultAcl;
```
PapaJoltOmega, line 737: why do we need default acl parameter?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
wiarlawd: This call site for getAcl isn't running the content SQL, but when called from isUserAuthorized it is. I don't think that exposing the conditional from getAclFromDb in isUserAuthorized feels right. Perhaps better would be to bury this check for aclSql != null in getAcl as well, so that both call sites just always call getAcl, which figures out how to do the work.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
wiarlawd: As I mentioned, I didn't entirely like the structure of getAclFromDb, but it has parallel methods for getDocFromDb and getStreamFromDb. All three methods just create a PreparedStatement, so a basic rename to prepare{Acl,Doc,Stream}Statement seems reasonable, but feels like a separate CL. Maybe we can finish this one and start a new pull request to massage the structure?

Also, these methods have functional utility for try-with-resources: we want to call prepareStatement and executeQuery but do something in between. So we either need these helper methods, or we need two nested try-with-resources.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
wiarlawd: Because the other call to buildAcl from getDocContent using the content SQL doesn't ever want to call rs.next(). We could check rsNext at the top and bottom of the loop in buildAcl, which is messier and another double-check, but it's in one method.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
wiarlawd: If we have aclSql but no ACL entries, we want an empty ACL to mark the document secure. But if we have no aclSql and no ACL entries, then we want a null ACL to mark the document public.

wiarlawd (commented):
Replies to PJ.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Yes, Done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
PapaJoltOmega: Ah, yes, that changed here https://github.com/googlegsa/database/commit/afe59e4949b506bf81e55c230448749dd9e54ab9#diff-dd03438283161f00e652d4769de767bf

I agree that some structure/renaming changes are fitting.  

I have made a client and am reading DatabaseAdaptor.java file to  re-acquaint myself with it. I plan to read it on Monday and make some suggestions then.

Thank you

PapaJoltOmega (commented):
Thanks for your thoughtful review John.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -817,7 +831,9 @@ private PreparedStatement getDocFromDb(Connection conn,
 
   private PreparedStatement getAclFromDb(Connection conn,
       String uniqueId) throws SQLException {
-    PreparedStatement st = conn.prepareStatement(aclSql);
+    // aclSql will only be null when called from isUserAuthorized.
+    String sql = (aclSql == null) ? singleDocContentSql : aclSql;
```
wiarlawd: There are two bugs here. We'll have to go back to the if statement, because as I suggested elsewhere, the UniqueKey method depends on the query being executed and not the data being read from the result. So if we use singleDocContentSql here, we need to call setContentSqlValues rather than setAclSqlValues. We could just call getDocFromDb in that case, but the logging would be odd, and I think it's worth being clear that we're executing this statement to fetch the ACL in this case. We need some tests for mismatches in singleDocContentSqlParameters and aclSqlParameters (note that everyDocIdSql must return all the columns needed for both, but it could, for example, return 2 columns, and have each of the other queries use a different column; to check for errors I think something like a varchar and an integer column might do the trick, or one query could use both columns and the other just one).

Second, we're actually using AllPublic rather than AccessChecker when aclSql is null (this is set in init). At the moment, aclSql really can't be null here, but that's a bug. We need to always check for ACLs in the AuthzAuthority, but AccessChecker needs some work. We need to deal with the two TODOs, and the fact that getAcl can return null (I think the latter is the key to the former: a null ACL makes a null identity acceptable, otherwise we need the identity to compare to the ACL). Obviously, we need some tests here as well.

wiarlawd (changes requested):

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: As discussed, changed the code and added tests.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -699,17 +689,38 @@ public static void main(String[] args) {
...
+        }
+      }
+    } else {
+      // Generate ACL if it came as part of result to singleDocContentSql.
+      return buildAcl(rs, aclPrincipalDelimiter, aclNamespace,
+          ResultSetNext.PROCESS_ONE_ROW, null);
+    }
+  }
+
+  private Acl processAclQuery(PreparedStatement st, Acl defaultAcl)
```
wiarlawd: This was a surprising choice. Why did you extract this method instead of leaving getAclFromDb? The latter is a parallel method with two siblings, so instead of that consistency, you added this one-off method, and I don't think it isolates concerns very well, or reads any cleaner (it checks aclSql for null 3 times in getAcl, for example).

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -293,11 +293,7 @@ public void init(AdaptorContext context) throws Exception {
       ignored.add("db.updateTimestampTimezone");
     }
 
-    if (aclSql == null) {
-      context.setAuthzAuthority(new AllPublic());
```
wiarlawd, line 295: AllPublic is unused now.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -1146,7 +1147,11 @@ private PreparedStatement getUpdateStreamFromDb(Connection conn)
       try (Connection conn = makeNewConnection()) {
         for (DocId id : ids) {
           log.log(Level.FINE, "about to get acl of doc {0}", id);
-          Acl acl = getAcl(conn, id.getUniqueId());
+          Acl acl = getAcl(conn, null, id.getUniqueId());
```
wiarlawd: You missed some changes here. Try running this code live by fetching public docs directly from the adaptor.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -699,17 +689,38 @@ public static void main(String[] args) {
...
+          ResultSetNext.PROCESS_ONE_ROW, null);
+    }
+  }
+
+  private Acl processAclQuery(PreparedStatement st, Acl defaultAcl)
+      throws SQLException {
+    try (ResultSet resSet = st.executeQuery()) {
       log.finer("got acl");
-      boolean hasResult = rs.next();
+      boolean hasResult = resSet.next();
```
wiarlawd: Why s/rs/resSet/? It's a non-standard abbreviation, and rs in a database connector seems perfectly clear.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -699,17 +689,38 @@ public static void main(String[] args) {
...
       if (!hasResult) {
-        // empty Acl ensures adaptor will mark this document as secure
-        return Acl.EMPTY;
+        return defaultAcl;
```
wiarlawd: This is a behavioral change, albeit a minor one. Why make this change?

srinicodebytes: When calling from isUserAuthorized and aclSql is null and has no results this needs to return null. Hence passing the defaultAcl and returning it.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Unintended change in this new method, and carried over from trying to make it two nested try-with-resources blocks.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
wiarlawd: In fact, it needs to not return null. I thought it was just a nuisance to return null (it will execute the content SQL rather than just skipping this search result), but it's actually a security hole. If the row doesn't exist, then it makes no sense to return a PERMIT for it.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Was trying to make it cleaner. Can't avoid aclSql check multiple times though. 
Using getAclFromDb(), the code would be

```
  private Acl getAcl(Connection conn, ResultSet rs, String uniqueId)
      throws SQLException {
    if (aclSql != null || rs == null) {
      try (PreparedStatement stmt = getAclFromDb(conn, uniqueId);
          ResultSet resSet = stmt.executeQuery()) {
        log.finer("got acl");
        boolean hasResult = resSet.next();
        if (!hasResult) {
          if (aclSql == null) {
            return null;
          } else {
            // empty Acl ensures adaptor will mark this document as secure
            return Acl.EMPTY;
          }
        } else {
          return buildAcl(resSet, aclPrincipalDelimiter, aclNamespace,
              ResultSetNext.PROCESS_ALL_ROWS, Acl.EMPTY);
        }
      }
    } else {
      // Generate ACL if it came as part of result to singleDocContentSql.
      return buildAcl(rs, aclPrincipalDelimiter, aclNamespace,
          ResultSetNext.PROCESS_ONE_ROW, null);
    }
  }
```

Restored it back.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Yes missed it, removed it.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2805,6 +2805,7 @@ public void testAuthzAuthority_public() throws Exception {
     moreEntries.put("db.everyDocIdSql", "select id from data");
     moreEntries.put("db.singleDocContentSql",
         "select * from data where id = ?");
+    executeUpdate("insert into data(id) values (1),(2)");
```
wiarlawd: This syntax is H2-specific and we eliminated it when porting the unit tests to other databases. What's the point of having two rows here, instead of just one? Also, why separate this from the CREATE TABLE? Next, note that this "public" test specifies a non-null identity. Finally, both the executeUpdate calls and the db.uniqueKey and db.singleDocContentSql properties are now specific to this test. In that case, we put the executeUpdate calls first, then create the moreEntries hash map and put db.singleDocContentSql into it.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2966,29 +2967,29 @@ public void testAuthzAuthoritySingleDocContentSql_noResults()
...
-    golden.put(new DocId("2"), AuthzStatus.PERMIT);
+    golden.put(new DocId("1"), AuthzStatus.INDETERMINATE);
+    golden.put(new DocId("2"), AuthzStatus.INDETERMINATE);
     assertEquals(golden, answers);
   }
 
   @Test
   public void testAuthzAuthoritySingleDocContentSql_permit() throws Exception {
-    executeUpdate("create table data(id integer, loginid varchar(20), "
+    executeUpdate("create table data(id integer, author varchar(20), "
```
wiarlawd: What's the s/loginid/author/g for?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -693,22 +689,36 @@ public static void main(String[] args) {
...
+        if (!hasResult) {
+          // empty Acl ensures adaptor will mark this document as secure
+          return Acl.EMPTY;
+        } else {
+          if (aclSql != null) {
+            return buildAcl(resSet, aclPrincipalDelimiter, aclNamespace,
+                ResultSetNext.PROCESS_ALL_ROWS, Acl.EMPTY);
+          } else {
+            return buildAcl(resSet, aclPrincipalDelimiter, aclNamespace,
+                ResultSetNext.PROCESS_ALL_ROWS, null);
```
wiarlawd: This should be PROCESS_ONE_ROW. We're back to 3 checks for aclSql == null, which I think suggests we need some refactoring. Maybe some kind of ACL strategy pattern, where we construct an appropriate object once for each scenario?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -693,22 +689,36 @@ public static void main(String[] args) {
...
-  private Acl getAcl(Connection conn, String uniqueId) throws SQLException {
-    try (PreparedStatement stmt = getAclFromDb(conn, uniqueId);
-        ResultSet rs = stmt.executeQuery()) {
-      log.finer("got acl");
-      return buildAcl(rs, aclPrincipalDelimiter, aclNamespace);
+  private Acl getAcl(Connection conn, ResultSet rs, String uniqueId)
+      throws SQLException {
+    if (aclSql != null || rs == null) {
+      try (PreparedStatement stmt = getAclFromDb(conn, uniqueId);
+          ResultSet resSet = stmt.executeQuery()) {
```
wiarlawd: resSet is a terrible name. It gives no indication of how it's different from rs, is uses a short name for a parameter and a longer name for a local variable, and it uses a non-standard abbreviation. Better would be something like s/rs/contentRs/g and then s/resSet/rs/g.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -693,22 +689,36 @@ public static void main(String[] args) {
...
+        log.finer("got acl");
+        boolean hasResult = resSet.next();
+        if (!hasResult) {
+          // empty Acl ensures adaptor will mark this document as secure
+          return Acl.EMPTY;
+        } else {
```
wiarlawd: else if

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -1109,32 +1125,33 @@ private PreparedStatement getUpdateStreamFromDb(Connection conn)
...
-          List<Acl> aclChain = Arrays.asList(acl);
-          log.log(Level.FINE,
-              "about to authorize user {0} for doc {1} and acl {2}",
-              new Object[]{user, id, acl});
-          AuthzStatus decision = Acl.isAuthorized(userIdentity, aclChain); 
-          log.log(Level.FINE,
-              "authorization decision {0} for user {1} and doc {2}",
-              new Object[]{decision, user, id});
-          result.put(id, decision);
+          if (userIdentity == null) {
```
wiarlawd: What did you hope to gain by just moving this check inside the loop? Did you try this live?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -1109,32 +1125,33 @@ private PreparedStatement getUpdateStreamFromDb(Connection conn)
...
+            result.put(id, AuthzStatus.DENY);
+          } else if (userIdentity.getUser() == null) {
+            result.put(id, AuthzStatus.DENY);
+          } else {
+            Acl acl = getAcl(conn, null, id.getUniqueId());
+            if (acl == null) {
+              result.put(id, AuthzStatus.PERMIT);
+            } else {
+              log.log(Level.INFO, "about to authorize {0} {1}",
```
wiarlawd: What this meant to stay here?

wiarlawd (changes requested):
Partial review. I didn't finish looking at the tests.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Yes, it should be PROCESS_ONE_ROW, thx. Done.
Yes, agree, refactoring to reduce the null check for aclSql will look cleaner.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Existing log statement moved.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: As discussed, changed to get and check Acl first.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Instead of id and loginid as unique keys, id and author sounds better. Hence changed it.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Only inserted data to existing test. Made changes as per comments.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -1109,32 +1123,29 @@ private PreparedStatement getUpdateStreamFromDb(Connection conn)
...
-              "authorization decision {0} for user {1} and doc {2}",
-              new Object[]{decision, user, id});
-          result.put(id, decision);
+          Acl acl = getAcl(conn, null, id.getUniqueId());
+          if (acl == null) {
+            result.put(id, AuthzStatus.PERMIT);
+          } else if (userIdentity == null || userIdentity.getUser() == null) {
+            result.put(id, AuthzStatus.DENY);
+          } else {
+            log.log(Level.INFO, "about to authorize {0} {1}",
```
wiarlawd: s/INFO/FINE/, since it's executed now for every document in the authZ request.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2711,23 +2797,48 @@ private AuthnIdentity getAuthnIdentity(final UserPrincipal user) {
...
+    // Required for validation, but not specific to this test.
+    moreEntries.put("db.modeOfOperation", "rowToText");
+    moreEntries.put("db.everyDocIdSql", "select id from data");
 
     AuthzAuthority authority = getAuthzAuthority(moreEntries);
     Map<DocId, AuthzStatus> answers = authority.isUserAuthorized(
-        getAuthnIdentity(new UserPrincipal("alice")),
-        Arrays.asList(new DocId("1"), new DocId("2")));
+        null,
+        Arrays.asList(new DocId("1")));
```
wiarlawd: Oh, I see now, there were two docs in this test. I somehow missed that. I'm OK with restoring that. It seems reasonable to maintain that and test multiple IDs.

wiarlawd (commented):
This looks right to me. Since we had to work so hard to get the parameters right, let's be sure to capture that in the tests. Some missing tests that I found:

* PROCESS_ONE_ROW for both calls (testSingleDocSqlResultSetHasTwoRecords is not doing the job for the contentRs case because the query only returns one row)

Previously untested behaviors that we changed the code paths for, including:

* PROCESS_ALL_ROWS (there are apparently no tests with multiple rows returned by aclSql, despite the existing code to process such)
* Acl.EMPTY (ditto: the implicit empty ACL the code used to return for what's now the Acl.EMPTY defaultAcl parameter is untested: an aclSql with no ACL columns)
* Combining aclSql with setContentSqlColumns (this was never happening, but I think it's reasonable to augment testAuthzAuthorityAcl_permit with your author field and swizzle things so that this combo would fail there)

I don't think this needs to be 5 new tests, maybe just some config or test data variations in existing tests could cover these.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Restored to maintain testing multiple IDs.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2809,27 +2923,28 @@ public void testAuthzAuthorityAcl_noResults() throws Exception {
   @Test
   public void testAuthzAuthorityAcl_permit() throws Exception {
     executeUpdate("create table acl(id integer, gsa_permit_users varchar(20))");
+    executeUpdate("create table data(id integer, author varchar(20))");
     executeUpdate("insert into acl(id, gsa_permit_users) values "
         + "(2, 'alice')");
 
     Map<String, String> moreEntries = new HashMap<String, String>();
     moreEntries.put("db.aclSql", "select * from acl where id = ?");
-    // Required for validation, but not specific to this test.
```
wiarlawd: Why did you remove this, or treat the data table, modeOfOperation, or everyDocIdSql as specific to this test? I can see moving db.uniqueKey and db.everyDocIdSql up (though I probably wouldn't have move the latter, since it's never executed).

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2860,4 +2975,135 @@ public void testAuthzAuthorityAcl_sqlException() throws Exception {
...
         Arrays.asList(new DocId("1"), new DocId("2")));
   }
+
+  @Test
+  public void testAuthzAuthoritySingleDocContentSql_noResults()
+      throws Exception {
+    executeUpdate("create table data(id integer)");
+
+    Map<String, String> moreEntries = new HashMap<String, String>();
+    moreEntries.put("db.modeOfOperation", "rowToText");
```
wiarlawd: Move to "Required for validation"

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2860,4 +2975,135 @@ public void testAuthzAuthorityAcl_sqlException() throws Exception {
...
+        + "gsa_permit_users varchar(20), gsa_deny_users varchar(20))");
+    executeUpdate("insert into data(id, author, gsa_deny_users) values "
+        + "(1, 'tom', 'alice')");
+    executeUpdate("insert into data(id, author, gsa_permit_users) values "
+        + "(2, 'alice', 'alice')");
+
+    Map<String, String> moreEntries = new HashMap<String, String>();
+    moreEntries.put("db.aclSqlParameters", "author");
+    moreEntries.put("db.singleDocContentSqlParameters", "id, author");
+    moreEntries.put("db.modeOfOperation", "rowToText");
```
wiarlawd: Move to "Required for validation"

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2809,27 +2923,28 @@ public void testAuthzAuthorityAcl_noResults() throws Exception {
   @Test
   public void testAuthzAuthorityAcl_permit() throws Exception {
     executeUpdate("create table acl(id integer, gsa_permit_users varchar(20))");
+    executeUpdate("create table data(id integer, author varchar(20))");
```
wiarlawd: This shouldn't be here at all, but elsewhere the pattern is "create data, insert data, create acl, insert acl".

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2860,4 +2975,135 @@ public void testAuthzAuthorityAcl_sqlException() throws Exception {
...
+  @Test
+  public void testAuthzAuthoritySingleDocContentSql_muiltiRowAcl()
+      throws Exception {
+    executeUpdate("create table data(id integer, "
+        + "gsa_permit_users varchar(20), gsa_deny_users varchar(20))");
+    executeUpdate("insert into data(id, gsa_deny_users) values (1, 'tom')");
+    executeUpdate("insert into data(id, gsa_permit_users) values (1, 'alice')");
+
+    Map<String, String> moreEntries = new HashMap<String, String>();
+    moreEntries.put("db.modeOfOperation", "rowToText");
```
wiarlawd: You see the pattern.

wiarlawd (commented):
Just some organizational nits in the tests.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: It looked odd to have "create table data.." after "moreEntries.put("db.aclSql"..", hence moved it to top. However, I see that the other tests have this pattern. Restored. it. Thx.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Done.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
srinicodebytes: Yes, thx. Moved it in all the new tests.

wiarlawd (commented):
Looks good. We're waiting a bit for PJ, and in the meantime, please do a "git merge master" locally and resolve the conflicts and push the commit here.

PR #114, opened 2017-12-16
First support for secondary metadata query.

aptls:
Add support for a secondary SQL query to retrieve metadata from a
second table. This first implementation supports a model in which
metadata is stored in a second table in one or more rows keyed by
some unique value from the original doc record. Three new config
parameters are used: db.extraMetadataSql (to specify the query),
db.extraMetadataSqlParameters (to list the column names whose
values are used as parameters in db.extraMetadataSql), and
optionally db.extraMetadataColumns (to specify the columns from
the metadata result set to use).

The addMetadata/addMetadataTo* methods were refactored a bit to
allow addMetadata to be used for the extraMetadata table as well.

The current implementation requires the key into the metadata
table to be part of the doc's unique key, but this is not
necessarily a requirement to support retrieving metadata from a
second table.

The config parameters and new methods use "extraMetadata" in
their names to identify them as applying to this secondary
query. The prior existence of "db.metadataColumns" as a parameter
applying to the primary doc query made new parameter names
starting with "db.metadata*" feel confusing. A better naming
scheme might be nice.

This implementation is expected to be extended with support for
other metadata structures, possibly in a manner similar to the
ResponseGenerator, so that should be taken into account when
thinking about config parameter names.

### DIFF
--- a/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
+++ b/src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
@@ -128,6 +128,8 @@ public class DatabaseAdaptor extends AbstractAdaptor {
   private String everyDocIdSql;
   private String singleDocContentSql;
   private MetadataColumns metadataColumns;
+  private String extraMetadataSql;
+  private MetadataColumns extraMetadataColumns;
   private ResponseGenerator respGenerator;
   private String aclSql;
   private String aclPrincipalDelimiter;
@@ -150,6 +152,9 @@ public class DatabaseAdaptor extends AbstractAdaptor {
     // when set to true, if "db.metadataColumns" is blank, it will use all
     // returned columns as metadata.
     config.addKey("db.includeAllColumnsAsMetadata", "false");
+    config.addKey("db.extraMetadataSql", "");
+    config.addKey("db.extraMetadataSqlParameters", "");
+    config.addKey("db.extraMetadataColumns", "");
     config.addKey("db.modeOfOperation", null);
     config.addKey("db.updateSql", "");
     config.addKey("db.aclSql", "");
@@ -234,6 +239,20 @@ public class DatabaseAdaptor extends AbstractAdaptor {
       }
     }
 
+    if (!isNullOrEmptyString(cfg.getValue("db.extraMetadataSql"))) {
+      extraMetadataSql = cfg.getValue("db.extraMetadataSql");
+      log.config("extra metadata sql: " + extraMetadataSql);
+      String extraMetadataColumnsConfig =
+          cfg.getValue("db.extraMetadataColumns");
+      if ("".equals(extraMetadataColumnsConfig)) {
+        extraMetadataColumns = null;
+        log.config("extra metadata columns: Use all columns in ResultSet");
+      } else {
+        extraMetadataColumns = new MetadataColumns(extraMetadataColumnsConfig);
+        log.config("extra metadata columns: " + extraMetadataColumns);
+      }
+    }
+
     modeOfOperation = cfg.getValue("db.modeOfOperation");
     log.config("mode of operation: " + modeOfOperation);
 
@@ -311,7 +330,9 @@ public class DatabaseAdaptor extends AbstractAdaptor {
         = new UniqueKey.Builder(cfg.getValue("db.uniqueKey"))
         .setDocIdIsUrl(docIdIsUrl)
         .setContentSqlColumns(cfg.getValue("db.singleDocContentSqlParameters"))
-        .setAclSqlColumns(cfg.getValue("db.aclSqlParameters"));
+        .setAclSqlColumns(cfg.getValue("db.aclSqlParameters"))
+        .setMetadataSqlColumns(
+            cfg.getValue("db.extraMetadataSqlParameters"));
 
     // Verify all column names.
     try (Connection conn = makeNewConnection()) {
@@ -331,6 +352,10 @@ public class DatabaseAdaptor extends AbstractAdaptor {
           verifyColumnNames(conn, "db.singleDocContentSql", singleDocContentSql,
               "db.metadataColumns", metadataColumns.keySet());
         }
+      }
+      if (extraMetadataColumns != null) {
+        verifyColumnNames(conn, "db.extraMetadataSql", extraMetadataSql,
+            "db.extraMetadataColumns", extraMetadataColumns.keySet());
       }
       if (respGenerator instanceof ResponseGenerator.SingleColumnContent) {
         ResponseGenerator.SingleColumnContent content =
@@ -477,6 +502,9 @@ public class DatabaseAdaptor extends AbstractAdaptor {
           builder.setDeleteFromIndex(true);
         } else if ("urlAndMetadataLister".equals(modeOfOperation)) {
           addMetadataToRecordBuilder(builder, rs);
+          if (extraMetadataSql != null) {
+            addExtraMetadataToRecordBuilder(builder, conn, id.getUniqueId());
+          }
         }
         DocIdPusher.Record record = builder.build();
         log.log(Level.FINEST, "doc id: {0}", id);
@@ -501,15 +529,10 @@ public class DatabaseAdaptor extends AbstractAdaptor {
   /*
    * Adds all specified metadata columns to the record or response being built.
    */
-  private void addMetadata(MetadataHandler meta, ResultSet rs)
-      throws SQLException, IOException {
+  private void addMetadata(MetadataHandler meta, ResultSet rs,
+      MetadataColumns rsMetadataColumns) throws SQLException, IOException {
     ResultSetMetaData rsMetaData = rs.getMetaData();
-    synchronized (this) {
-      if (metadataColumns == null) {
-        metadataColumns = new MetadataColumns(rsMetaData);
-      }
-    }
-    for (Map.Entry<String, String> entry : metadataColumns.entrySet()) {
+    for (Map.Entry<String, String> entry : rsMetadataColumns.entrySet()) {
       int index;
       try {
         index = rs.findColumn(entry.getKey());
@@ -633,25 +656,81 @@ public class DatabaseAdaptor extends AbstractAdaptor {
   @VisibleForTesting
   void addMetadataToRecordBuilder(final DocIdPusher.Record.Builder builder,
       ResultSet rs) throws SQLException, IOException {
+    synchronized (this) {
+      if (metadataColumns == null) {
+        metadataColumns = new MetadataColumns(rs.getMetaData());
+      }
+    }
     addMetadata(
         new MetadataHandler() {
           @Override public void addMetadata(String k, String v) {
             builder.addMetadata(k, v);
           }
         },
-        rs);
+        rs, metadataColumns);
   }
 
   @VisibleForTesting
   void addMetadataToResponse(final Response resp, ResultSet rs)
       throws SQLException, IOException {
+    synchronized (this) {
+      if (metadataColumns == null) {
+        metadataColumns = new MetadataColumns(rs.getMetaData());
+      }
+    }
     addMetadata(
         new MetadataHandler() {
           @Override public void addMetadata(String k, String v) {
             resp.addMetadata(k, v);
           }
         },
-        rs);
+        rs, metadataColumns);
+  }
+
+  private void addExtraMetadataToRecordBuilder(
+      final DocIdPusher.Record.Builder builder, Connection conn,
+      String uniqueId) throws SQLException, IOException {
+
+    try (PreparedStatement stmt = getExtraMetadataFromDb(conn, uniqueId);
+        ResultSet rs = stmt.executeQuery()) {
+      synchronized (this) {
+        if (extraMetadataColumns == null) {
+          extraMetadataColumns = new MetadataColumns(rs.getMetaData());
+        }
+      }
+      MetadataHandler handler = new MetadataHandler() {
+          @Override public void addMetadata(String k, String v) {
+            builder.addMetadata(k, v);
+          }
+        };
+      while (rs.next()) {
+        addMetadata(handler, rs, extraMetadataColumns);
+      }
+    } catch (SQLException ex) {
+      throw new IOException(ex);
+    }
+  }
+
+  private void addExtraMetadataToResponse(final Response resp, Connection conn,
+      String uniqueId) throws SQLException, IOException {
+    try (PreparedStatement stmt = getExtraMetadataFromDb(conn, uniqueId);
+        ResultSet rs = stmt.executeQuery()) {
+      synchronized (this) {
+        if (extraMetadataColumns == null) {
+          extraMetadataColumns = new MetadataColumns(rs.getMetaData());
+        }
+      }
+      MetadataHandler handler = new MetadataHandler() {
+          @Override public void addMetadata(String k, String v) {
+            resp.addMetadata(k, v);
+          }
+        };
+      while (rs.next()) {
+        addMetadata(handler, rs, extraMetadataColumns);
+      }
+    } catch (SQLException ex) {
+      throw new IOException(ex);
+    }
   }
 
   /** Gives the bytes of a document referenced with id. */
@@ -675,6 +754,9 @@ public class DatabaseAdaptor extends AbstractAdaptor {
       }
       // Generate response metadata first.
       addMetadataToResponse(resp, rs);
+      if (extraMetadataSql != null) {
+        addExtraMetadataToResponse(resp, conn, id.getUniqueId());
+      }
       // Generate Acl if aclSql is provided.
       if (aclSql != null) {
         resp.setAcl(getAcl(conn, id.getUniqueId()));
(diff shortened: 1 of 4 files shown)


src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -234,6 +239,20 @@ public void init(AdaptorContext context) throws Exception {
       }
     }
 
+    if (!isNullOrEmptyString(cfg.getValue("db.extraMetadataSql"))) {
```
wiarlawd: This can't be null; we've been drawing the conditions more narrowly in new code. I prefer using foo.isEmpty, but "".equals(foo) is popular in the code, too.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -234,6 +239,20 @@ public void init(AdaptorContext context) throws Exception {
...
+      String extraMetadataColumnsConfig =
+          cfg.getValue("db.extraMetadataColumns");
+      if ("".equals(extraMetadataColumnsConfig)) {
+        extraMetadataColumns = null;
+        log.config("extra metadata columns: Use all columns in ResultSet");
+      } else {
+        extraMetadataColumns = new MetadataColumns(extraMetadataColumnsConfig);
+        log.config("extra metadata columns: " + extraMetadataColumns);
+      }
+    }
```
wiarlawd: Add an else to check for non-empty db.extraMetadataColumns and db.extraMetadataSqlParameters and add them to ignored. You should add a test for that as well (I think ignoring both in one test is fine, or maybe even add it to an existing ignored property test; I'm not sure how those are organized.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -501,15 +529,10 @@ private boolean isDeleteAction(ResultSet rs) throws SQLException {
   /*
    * Adds all specified metadata columns to the record or response being built.
    */
-  private void addMetadata(MetadataHandler meta, ResultSet rs)
-      throws SQLException, IOException {
+  private void addMetadata(MetadataHandler meta, ResultSet rs,
+      MetadataColumns rsMetadataColumns) throws SQLException, IOException {
```
wiarlawd: Maybe just columns? rsMetadataColumns made me expect a ResultSetMetaData object.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -633,25 +656,81 @@ private void addMetadata(MetadataHandler meta, ResultSet rs)
...
           }
         },
-        rs);
+        rs, metadataColumns);
+  }
+
+  private void addExtraMetadataToRecordBuilder(
+      final DocIdPusher.Record.Builder builder, Connection conn,
+      String uniqueId) throws SQLException, IOException {
+
```
wiarlawd: Delete blank line.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -633,25 +656,81 @@ private void addMetadata(MetadataHandler meta, ResultSet rs)
...
+    try (PreparedStatement stmt = getExtraMetadataFromDb(conn, uniqueId);
```
wiarlawd: What made you decide against the extra helper method to share this try statement between the two new methods?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -838,6 +920,14 @@ private PreparedStatement getStreamFromDb(Connection conn,
     return st;
   }
 
+  private PreparedStatement getExtraMetadataFromDb(Connection conn,
```
wiarlawd: We're planning to rename these methods, so lets go with prepareExtraMetadataStatement for this one now, so the change doesn't collide with this one.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -217,8 +224,8 @@ static String decodeSlashInData(String id) {
 
   @Override
   public String toString() {
-    return "UniqueKey(" + docIdSqlCols + "," + types + "," + contentSqlCols + ","
-        + aclSqlCols + ")";
+    return "UniqueKey(" + docIdSqlCols + "," + types + ","
+        + contentSqlCols + "," + aclSqlCols + "," + metadataSqlCols + ")";
```
wiarlawd: Why re-wrap this and modify the first line? Random or on purpose?

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2643,6 +2682,124 @@ public void testMetadataColumns_columnAlias() throws Exception {
...
+  public void testExtraMetadata_docAndExtra() throws Exception {
+    executeUpdate("create table data(id int, name varchar(200))");
+    executeUpdate("create table metadata(id int, other varchar(200))");
+    executeUpdate("insert into data(id, name) values (1001, 'Doe, John')");
+    executeUpdate("insert into metadata(id, other) values (1001, 'John 1')");
+    executeUpdate("insert into metadata(id, other) values (1001, 'John 2')");
+
+    Map<String, String> configEntries = new HashMap<String, String>();
+    configEntries.put("db.uniqueKey", "id:int");
+    configEntries.put("db.everyDocIdSql", "select * from data");
```
wiarlawd: Unused, here and the other new tests, so put it after the usual comment to that effect.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2643,6 +2682,124 @@ public void testMetadataColumns_columnAlias() throws Exception {
...
+    configEntries.put("db.singleDocContentSql",
+        "select id, name from data where id = ?");
+    configEntries.put("db.modeOfOperation", "rowToText");
+    // extraMetadataSql uses "select *", so extraMetadataColumns
+    // is used to limit the columns returned as metadata
+    configEntries.put("db.extraMetadataSql",
+        "select * from metadata where id = ?");
+    configEntries.put("db.extraMetadataSqlParameters", "id");
+    configEntries.put("db.extraMetadataColumns", "OTHER");
```
wiarlawd, line 3018: I think we could either leave this lowercase, or maybe mixed case? Or, since I think we should have a test with a mapping in db.extraMetadataColumns, we could use uppercase in one, as here, and lower case in the other, as "other:col1".

wiarlawd (changes requested):
Lots of comments but mostly minor suggestions or questions.

Check the code coverage. It looks like you're missing tests for urlAndMetadataLister, including both getDocIds and getModifiedDocIds.

src/com/google/enterprise/adaptor/database/UniqueKey.java
aptls: The line was triggering my 80-cols check and since I was adding to the string anyway, I changed it to get rid of the warning.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -633,25 +672,73 @@ private void addMetadata(MetadataHandler meta, ResultSet rs)
...
           }
         },
-        rs);
+        rs, metadataColumns);
+  }
+
+  private void addExtraMetadataToRecordBuilder(
+      final DocIdPusher.Record.Builder builder, Connection conn,
+      String uniqueId) throws SQLException, IOException {
+    MetadataHandler handler = new MetadataHandler() {
```
wiarlawd: Maybe inline this as we do in addMetadataToXxx?

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -218,7 +229,7 @@ static String decodeSlashInData(String id) {
   @Override
   public String toString() {
     return "UniqueKey(" + docIdSqlCols + "," + types + "," + contentSqlCols + ","
-        + aclSqlCols + ")";
+        + aclSqlCols + "," + metadataSqlCols + ")";
```
wiarlawd, line 244: Har, OK. Either way is fine.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -345,6 +357,19 @@ Builder setAclSqlColumns(String aclSqlColumns)
       return this;
     }
 
+    Builder setMetadataSqlColumns(String metadataSqlColumns)
+        throws InvalidConfigurationException {
+      if (null == metadataSqlColumns) {
```
wiarlawd: Use Yoda conditions, we do not. (Used in the past, they have been, but no longer.)

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -1145,6 +1171,28 @@ public void testInitLister_ignoredProperties_allColumns() throws Exception {
         containsString("[db.includeAllColumnsAsMetadata]"));
    }
 
+  @Test
+  public void testInit_ignoredPropertyes_extraMetadata() throws Exception {
```
wiarlawd: s/Propertyes/Properties/, unless these are specifically ye olde style propertyes.

Also, I think testInitExtraMetadata_..., for consistency with the other names (we aren't using testInit_... anywhere, though perhaps that would be better).

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -114,18 +116,22 @@ String makeUniqueId(ResultSet rs) throws SQLException, URISyntaxException {
...
-      String errmsg = "Wrong number of values for primary key: "
-          + "id: " + uniqueId + ", parts: " + Arrays.asList(parts);
-      throw new IllegalStateException(errmsg);
-    }
     Map<String, String> zip = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
-    for (int i = 0; i < parts.length; i++) {
-      String columnValue = decodeSlashInData(parts[i]);
-      zip.put(docIdSqlCols.get(i), columnValue);
+    if (docIdIsUrl) {
+      zip.put(docIdSqlCols.get(0), uniqueId);
```
wiarlawd: I don't like this. It feels so special case: if your doc ID is an URL (whether using urlAndMetadataLister mode or not), then your extraMetadataSql can only take that URL as a unique key parameter. We should make that explicit somewhere, maybe in a comment here. As we discussed, I think it would be preferable to try the route you suggested originally where the extraMetadataSqlParameters come from the singleDocContentSql result set, rather than from the doc ID. Maybe just a TODO here for now, I'm OK with exploring that in a separate change, but PJ may not be.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -633,25 +672,73 @@ private void addMetadata(MetadataHandler meta, ResultSet rs)
...
-        rs);
+        rs, metadataColumns);
   }
 
   @VisibleForTesting
   void addMetadataToResponse(final Response resp, ResultSet rs)
       throws SQLException, IOException {
+    synchronized (this) {
+      if (metadataColumns == null) {
+        metadataColumns = new MetadataColumns(rs.getMetaData());
```
wiarlawd, line 694: This line isn't tested, since you lifted this out of the helper in two places, only the other line is executed.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -633,25 +672,73 @@ private void addMetadata(MetadataHandler meta, ResultSet rs)
...
+      synchronized (this) {
+        if (extraMetadataColumns == null) {
+          extraMetadataColumns = new MetadataColumns(rs.getMetaData());
+        }
+      }
+      while (rs.next()) {
+        addMetadata(handler, rs, extraMetadataColumns);
+      }
+    } catch (SQLException ex) {
+      throw new IOException(ex);
```
wiarlawd, line 740: This line is untested. I'm not sure how to trigger this. A flat invalid query would fail in validation. Maybe a type mismatch in the parameter expression: `<int-field> = ?` with a character parameter? I can look harder if you can't find something easily.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -1404,6 +1452,45 @@ public void testInitVerifyColumnNames_metadataColumnAliasBad()
...
+  public void testInitVerifyColumnNames_metadataSql_columnInUniqueKey()
+      throws Exception {
+    executeUpdate("create table data(id int, other varchar(20))");
+
+    Map<String, String> moreEntries = new HashMap<String, String>();
+    moreEntries.put("db.modeOfOperation", "rowToText");
+    moreEntries.put("db.uniqueKey", "id:int");
+    moreEntries.put("db.extraMetadataSql",
+        "select other from data where id = ?");
+    moreEntries.put("db.extraMetadataSqlParameters", "id");
```
wiarlawd, line 1465: Do you have a test that does not specify this? By default, db.uniqueKey should be used, so we should have a test of that, which could maybe be this one. I think we should have some test where the unique key has two columns, one used in singleDocContentSql, and the other used in extraMetadataSql.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2643,6 +2812,161 @@ public void testMetadataColumns_columnAlias() throws Exception {
...
+    RecordingResponse response = new RecordingResponse();
+    adaptor.getDocContent(request, response);
+    Metadata expected = new Metadata();
+    expected.add("OTHER", "John 1");
+    expected.add("OTHER", "John 2");
+    assertEquals(expected, response.getMetadata());
+  }
+
+  @Test
+  public void testExtraMetadata_extraMetadataColumns_Mapping()
```
wiarlawd: s/Mapping/mapping/

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -1527,6 +1614,44 @@ public void testGetDocIds_urlAndMetadataLister() throws Exception {
         pusher.getRecords());
   }
 
+  @Test
```
wiarlawd, line 1617: Another test that I don't see from before that we could have had, is to make sure that getDocIds and getModifiedDocIds really do ignore the metadata config when docIdIsUrl. We have lots of init tests, and one each for getDocIds and getModifiedDocIds, to check the URL and date handling. I think we should either add some metadata to those, or have a new pair of tests with just a single row, and both metadataColumns and extraMetadataSql configured, and verify that no metadata is returned (and maybe extraMetadataSql could be syntactically invalid; though I think we "ignore" the property, but still validate the SQL?).

wiarlawd (changes requested):
Mostly ideas for more tests, and the issue with docId.isUrl and db.extraMetadataSql.

src/com/google/enterprise/adaptor/database/UniqueKey.java
```diff
@@ -117,6 +117,15 @@ String makeUniqueId(ResultSet rs) throws SQLException, URISyntaxException {
   private void setSqlValues(PreparedStatement st, String uniqueId,
       List<String> sqlCols) throws SQLException {
     Map<String, String> zip = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+    // Parameters to SQL queries are taken from the unique id. When reading extraMetadata
```
wiarlawd: 80 cols

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -1686,6 +1727,7 @@ public void testGetDocIds_docIdIsUrl() throws Exception {
         pusher.getRecords());
   }
 
+
```
wiarlawd: Delete added blank line.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
```diff
@@ -2825,7 +2916,7 @@ public void testExtraMetadata_docAndExtra() throws Exception {
     configEntries.put("db.singleDocContentSql",
         "select id, name from data where id = ?");
     configEntries.put("db.modeOfOperation", "rowToText");
-    configEntries.put("db.metadataColumns", "name");
+    configEntries.put("db.includeAllColumnsAsMetadata", "true");
```
wiarlawd: Why make this change?

wiarlawd (approved):
Still waiting for PJ's review.

test/com/google/enterprise/adaptor/database/DatabaseAdaptorTest.java
aptls: Changing this setting adds coverage for the case you noted above for line 694.

PR #115, opened 2017-12-28
Fix bug# 67471678 - Support If-Modified-Since.

srinicodebytes:


src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -462,6 +485,8 @@ public void getDocIds(DocIdPusher pusher) throws IOException,
       log.finer("queried for stream");
       boolean hasAction
           = hasColumn(rs.getMetaData(), GsaSpecialColumns.GSA_ACTION);
+      boolean hasTimestamp =
```
wiarlawd: This is in between getting hasAction and logging hasAction, and we're not logging hasTimestamp. Either get/log/get/log or get/get/log/log.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -472,6 +497,20 @@ public void getDocIds(DocIdPusher pusher) throws IOException,
...
           continue;
         }
+        // Cache last modified time stamp for this DocId
+        if (hasTimestamp) {
+          Timestamp lastDocIdTimestamp = docIdLastModifiedMap.getIfPresent(id);
+          Timestamp ts =
+              rs.getTimestamp(GsaSpecialColumns.GSA_TIMESTAMP.toString(),
+                  updateTimestampTimezone);
+          if ((ts != null)
+              && (lastDocIdTimestamp == null || ts.after(lastDocIdTimestamp))) {
```
wiarlawd: What's the merit of checking the cache here, and what's a plausible scenario where ts < lastDocIdTimestamp?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -78,6 +80,9 @@
   /** Map from SQL types to SQL type names. */
   private static final HashMap<Integer, String> sqlTypeNames = new HashMap<>();
 
+  /** Cache for DocId and last modified time stamp. */
+  private static Cache<DocId, Timestamp> docIdLastModifiedMap;
```
wiarlawd: s/docId//; I know it's a map from DocId to Timestamp, but it's really a cache of last modified dates. The "docId" feels like noise. I could see using "Timestamp" consistently, since that's the name of the column, so maybe "lastModifiedCache" or "timestampCache"?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -673,6 +712,15 @@ public void getDocContent(Request req, Response resp) throws IOException {
         resp.respondNotFound();
         return;
       }
+      // Check if modified since last access.
+      Timestamp lastDocIdTimestamp = docIdLastModifiedMap.getIfPresent(id);
+      if (lastDocIdTimestamp != null
+          && !req.hasChangedSinceLastAccess(
+              new Date(lastDocIdTimestamp.getTime()))) {
```
wiarlawd: You should store Date rather than timestamp, to avoid having to do this for each crawl request. That also avoids the need to check for null here (because hasChangedSinceLastAccess does that, too).

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -472,6 +497,20 @@ public void getDocIds(DocIdPusher pusher) throws IOException,
               + e.getMessage());
           continue;
         }
+        // Cache last modified time stamp for this DocId
+        if (hasTimestamp) {
+          Timestamp lastDocIdTimestamp = docIdLastModifiedMap.getIfPresent(id);
```
wiarlawd: What does "last" mean here? I think either "cachedTimestamp" or "cachedLastModified".

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -673,6 +712,15 @@ public void getDocContent(Request req, Response resp) throws IOException {
         resp.respondNotFound();
         return;
       }
+      // Check if modified since last access.
```
wiarlawd: This is supposed to happen before the query, or the cache is silly, and we could just compare it to a GSA_TIMESTAMP column in these results (which would be a valid design choice, but the whole idea was to avoid hitting the database in case that's unusually expensive).

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -682,7 +730,8 @@ public void getDocContent(Request req, Response resp) throws IOException {
       // Generate response body.
       // In database adaptor's case, we almost never want to follow the URLs.
       // One record means one document.
-      resp.setNoFollow(true); 
+      log.log(Level.FINE, "Content modified since last crawl: {0}", id);
+      resp.setNoFollow(true);
```
wiarlawd: Whitespace only change.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -682,7 +730,8 @@ public void getDocContent(Request req, Response resp) throws IOException {
       // Generate response body.
       // In database adaptor's case, we almost never want to follow the URLs.
       // One record means one document.
-      resp.setNoFollow(true); 
+      log.log(Level.FINE, "Content modified since last crawl: {0}", id);
```
wiarlawd: Why not do this right away, on line 724?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -136,6 +141,9 @@ private static boolean isNullOrEmptyString(String str) {
   private boolean docIdIsUrl;
   private String modeOfOperation;
 
+  private Calendar updateTimestampTimezone;
+  private DateFormat formatter;
```
wiarlawd: You've got a threading problem here. getDocIds and getModifiedDocIds can run concurrently (though each one separately cannot). We could 1) maintain a separate formatter for each of them (slightly tedious but not a bad option), 2) synchronize the uses, 3) put these into a ThreadLocal (boo), or 4) create an instance for each call (again, slightly tedious, but not a bad option, given that calls happen every 15 minutes at best, by default). I'm leaning toward option 1.

wiarlawd (changes requested):
Partial review. I haven't looked at the tests yet.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -345,6 +365,9 @@ public void init(AdaptorContext context) throws Exception {
 
     uniqueKey = ukBuilder.build();
     log.config("primary key: " + uniqueKey);
+
+    docIdLastModifiedMap =
+        CacheBuilder.newBuilder().maximumSize(1000000).build();
```
wiarlawd: We might want to set an initial capacity. The (undocumented) default is 16 (though I think it might be 40 in practice), which is pretty small, and will lead to lots of allocations to grow the cache (done by doubling). Maybe initialCapacity(10000)?

wiarlawd: maximumSize takes a long, so add L to the value (FYI, initialCapacity takes an int).

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -462,6 +485,8 @@ public void getDocIds(DocIdPusher pusher) throws IOException,
       log.finer("queried for stream");
       boolean hasAction
           = hasColumn(rs.getMetaData(), GsaSpecialColumns.GSA_ACTION);
```
wiarlawd: Should be OK to get the ResultSetMetaData twice, but let's not.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
wiarlawd: I thought we were going to clear the cache in getDocIds (but not getModifiedDocIds, obviously). That avoids keeping a record for deleted rows, and in particular, if the ordering in getDocContent is fixed, avoids us reporting deleted rows as unmodified instead.

wiarlawd (commented):
More thoughts, still partial.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Thx, meant to change the name from "..Map" to "..Cache". Removed "docId" prefix as suggested.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Thx, done.

srinicodebytes: Thx, done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Ok, done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done.

srinicodebytes: Yes, missed it. Added a call to invalidateAll().

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Trying to check if the ts is newer than cached value before updating cache. ts won't be < lastDocIdTimestamp, but can be same as lastDocIdTimestamp. The additional check is not required if I update the cache always..
 
Removed the check.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Moved to line 724.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Renamed to "cachedLastModified"

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Restored the whitespace

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Yes, moved it.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Changed to have separate formatter for each (option 1).
Done.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -478,16 +476,19 @@ public String getName() {
   @Override
   public void getDocIds(DocIdPusher pusher) throws IOException,
       InterruptedException {
+    DateFormat formatter = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS Z");
+    formatter.setTimeZone(updateTimestampTimezone.getTimeZone());
     BufferedPusher outstream = new BufferedPusher(pusher);
+    lastModifiedCache.invalidateAll();
```
wiarlawd: I think I would move this to the first line (it's tied to the semantics of getDocIds, not this try statement or the loop, and these fields/variables are used in this order below: lastModifiedCache, formatter, outstream).

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -499,16 +500,15 @@ public void getDocIds(DocIdPusher pusher) throws IOException,
         }
         // Cache last modified time stamp for this DocId
         if (hasTimestamp) {
-          Timestamp lastDocIdTimestamp = docIdLastModifiedMap.getIfPresent(id);
+          Date cachedLastModified = lastModifiedCache.getIfPresent(id);
```
wiarlawd: This is never used, now, is it? Just overridden below and logged?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -499,16 +500,15 @@ public void getDocIds(DocIdPusher pusher) throws IOException,
...
           Timestamp ts =
               rs.getTimestamp(GsaSpecialColumns.GSA_TIMESTAMP.toString(),
                   updateTimestampTimezone);
-          if ((ts != null)
-              && (lastDocIdTimestamp == null || ts.after(lastDocIdTimestamp))) {
-            lastDocIdTimestamp = ts;
-            docIdLastModifiedMap.put(id, lastDocIdTimestamp);
-            log.log(Level.FINE, "docIdLastModifiedMap updated: {0}",
-                formatter.format(new Date(lastDocIdTimestamp.getTime())));
+          if ((ts != null)) {
```
wiarlawd: Delete extra parens.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -1081,11 +1081,11 @@ public void getModifiedDocIds(DocIdPusher pusher)
                 log.log(Level.FINE, "latestTimestamp updated: {0}",
                     formatter.format(new Date(latestTimestamp.getTime())));
               }
-              if (lastDocIdTimestamp == null || ts.after(lastDocIdTimestamp)) {
-                lastDocIdTimestamp = ts;
-                docIdLastModifiedMap.put(id, lastDocIdTimestamp);
-                log.log(Level.FINE, "lastDocIdTimestamp updated: {0}",
-                    formatter.format(new Date(lastDocIdTimestamp.getTime())));
+              if (cachedLastModified == null) {
```
wiarlawd: This if statement is wrong. Why would a row have to be uncached in order to update the cache with a new last modified?

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -1081,11 +1081,11 @@ public void getModifiedDocIds(DocIdPusher pusher)
...
+                cachedLastModified = new Date(ts.getTime());
```
wiarlawd: We're constructing this same Date object twice here, maybe they could be shared, if it doesn't seem too artificial.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
```diff
@@ -478,16 +476,19 @@ public String getName() {
   @Override
   public void getDocIds(DocIdPusher pusher) throws IOException,
       InterruptedException {
+    DateFormat formatter = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS Z");
```
wiarlawd: This is option 4, not option 1, but that's fine.

wiarlawd (commented):
I still haven't read the tests, but you changed the semantics of the prod code without changing any tests, which isn't a good sign.

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Yes, removed call to getIfPresent()

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done, thx

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Done

src/com/google/enterprise/adaptor/database/DatabaseAdaptor.java
srinicodebytes: Fixed it.

srinicodebytes (commented):
Code changes were mostly around updating cache. Earlier there were additional checks and cache is updated only if newer, now cache is always updated. Hence the tests didn't change. New tests are mostly around null modified date,  modified date is newer or not modified.

PR #117, opened 2018-02-27
db.modeOfOperation optional when isUrl = true (b/71554204)

wiarlawd:
When docid.isUrl is true, getDocContent returns NOT_FOUND immediately,
and the ResponseGenerator is never used, so it need not be configured.

The ResponseGenerator for urlAndMetadataLister was never needed, for
the same reason. I left it in place, since it's a public method in a
public class.


PapaJoltOmega (approved):
Thank you LGTM

